PR #109, opened 2014-12-16
Code in ProducerInjector duplicated in AsyncProducerInjector 

alex-on-java:
Three private methods (makeSpecialized, updateQualifiersAndName and methodSignatureMatches) was completely equal in ProducerInjector and AsyncProducerInjector classes.

I tried to put these method to their common ancestor, but code becomes uglier than it was, so I moved them to separate class.

As I'm new to this project, I don't know where is it a good idea, but I hope this change will make it easier to maintain errai.

P.S. May be names of new class & interface are not good, please change them if you construct better ones.


mbarkley:
There is a new implementation of Errai IoC on [this branch](https://github.com/errai/errai/tree/4.0) that will soon replace the one on master, at which point this will be impossible to merge. Because of this and because there has been no activity here for several months, I am closing this PR.

PR #118, opened 2015-02-15
ERRAI-841: Add support for non safe url chars in path parameters.

iliya-gr:


divd:
Excellent, thank you very much!

PR #119, opened 2015-02-18
Improvements on JBossLauncher usability (basically new defaults)

porcelli:
With these changes it's way easier to setup JBossLauncher from within IDE's.


csadilek:
After discussions we're closing this in favour of the new [EmbeddedWildFlyLauncher] (https://github.com/errai/errai/blob/master/errai-cdi/jboss/src/main/java/org/jboss/errai/cdi/server/gwt/EmbeddedWildFlyLauncher.java) which doesn't need a JavaAgent for hiding client-local classes.

PR #120, opened 2015-02-19
[ERRAI-841] Fix URLPattern url generation.

iliya-gr:
- Update printURL to match URLPatternMatcher generatePattern method
- Remove URL.decodePathSegment from tests
- Add test case without extra state param


divd:
Hello,

I took a look at your pull request, and it actually led me to a cleaner solution to the non-ASCII URL problem, while also solving the double encoding issue you mentioned in the JIRA. The changes should be available on 3.2.0-SNAPSHOT in 3-4 hours. We would appreciate it if you could test the new snapshots with your code and let us know if it fixes your problem. 

Thank you once again for your help.

PR #121, opened 2015-02-21
check if user has RequiredRoles and delegating the query to PicketLink

bauna:
The main idea of this change is to check if a user has the Roles asking that to PicketLink.
The previous version only checked if the user was directly granted a role, but if the user has the role for being a member of a group has the role the client received no roles.

### DIFF
--- a/errai-security/errai-security-picketlink/src/main/java/org/jboss/errai/security/server/PicketLinkAuthenticationService.java
+++ b/errai-security/errai-security-picketlink/src/main/java/org/jboss/errai/security/server/PicketLinkAuthenticationService.java
@@ -30,8 +30,8 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
 import org.jboss.errai.bus.server.annotations.Service;
+import org.jboss.errai.security.shared.api.RequiredRolesProvider;
 import org.jboss.errai.security.shared.api.Role;
-import org.jboss.errai.security.shared.api.RoleImpl;
 import org.jboss.errai.security.shared.api.identity.User;
 import org.jboss.errai.security.shared.api.identity.UserImpl;
 import org.jboss.errai.security.shared.exception.AlreadyLoggedInException;
@@ -42,11 +42,8 @@ import org.picketlink.Identity;
 import org.picketlink.Identity.AuthenticationResult;
 import org.picketlink.authentication.UserAlreadyLoggedInException;
 import org.picketlink.credential.DefaultLoginCredentials;
-import org.picketlink.idm.RelationshipManager;
 import org.picketlink.idm.credential.Password;
 import org.picketlink.idm.model.Attribute;
-import org.picketlink.idm.model.basic.Grant;
-import org.picketlink.idm.query.RelationshipQuery;
 
 /**
  * PicketLink version of the AuthenticationService and default implementation.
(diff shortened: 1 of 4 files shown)


errai-security/errai-security-picketlink/src/main/java/org/jboss/errai/security/server/PicketLinkBasicModelServicesImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.picketlink.idm.IdentityManagementException;
+import org.picketlink.idm.IdentityManager;
+import org.picketlink.idm.RelationshipManager;
+import org.picketlink.idm.model.IdentityType;
+import org.picketlink.idm.model.basic.BasicModel;
+import org.picketlink.idm.model.basic.Role;
+
+@Dependent
+public class PicketLinkBasicModelServicesImpl implements PicketLinkBasicModelServices {
```
bauna: Delegate implementation to BasicModel

errai-security/errai-security-picketlink/src/main/java/org/jboss/errai/security/server/PicketLinkBasicModelServices.java
```diff
@@ -0,0 +1,66 @@
+package org.jboss.errai.security.server;
+
+import org.picketlink.idm.IdentityManagementException;
+import org.picketlink.idm.model.Account;
+import org.picketlink.idm.model.IdentityType;
+import org.picketlink.idm.model.basic.Group;
+import org.picketlink.idm.model.basic.Role;
+
+public interface PicketLinkBasicModelServices {
```
bauna: Since BasicModel class from PicketLink has all method static, I created this interface to be able of mocking it in the Tests.

errai-security/errai-security-picketlink/src/main/java/org/jboss/errai/security/server/PicketLinkAuthenticationService.java
```diff
@@ -140,11 +140,10 @@ public User getUser() {
     Set<Role> roles = new HashSet<Role>();
 
     if (identity.isLoggedIn()) {
```
bauna: below is the main change of this PR.
The main difference is that now it ask PicketLink if the user has a Role.

errai-security/errai-security-picketlink/src/main/java/org/jboss/errai/security/server/PicketLinkAuthenticationService.java
```diff
@@ -61,11 +58,14 @@
...
   @Inject
-  private RelationshipManager relationshipManager;
+  private PicketLinkBasicModelServices basicModelServices;
 
   @Inject
   private DefaultLoginCredentials credentials;
 
+  @Inject
+  private RequiredRolesProvider requiredRolesProvider;
+
```
csadilek: Unfortunately, injecting a RequiredRolesProvider here won't work. It was mainly designed for the client so roles can be provided at runtime, see RestrictedAccess(provider=...). In fact, this will actually cause deployment errors for all applications that don't provide an implementation of that interface! We definitely do not want that.

csadilek:
This looks good in theory, but we need a different way to query the current user's roles. See my comment above. Thanks!

PR #123, opened 2015-02-26
[ERRAI-841] Fix URLPattern url generation.

iliya-gr:
- Update printURL to match URLPatternMatcher generatePattern method
- Remove URL.decodePathSegment from tests
- Add test case without extra state param
- Cleanup code
- Add test case where param has '/'
- Disable HistoryImplPushState encoding/decoding


errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/pushstate/HistoryImplPushState.java
```diff
@@ -66,7 +66,7 @@ public boolean init() {
   }
   
   private void nativeUpdate(final String historyToken) {
-    String newPushStateToken = CodeServerParameterHelper.append(encodeFragment(historyToken));
```
iliya-gr: I am not absolutely sure in removing of encodeFragment and decodeFragment.

divd:
Please see my reply at https://issues.jboss.org/browse/ERRAI-841 regarding this pull request.

PR #127, opened 2015-03-13
Bug 126 3.1.2.final

ManuelB:
This is a fix for the java function limit.


csadilek:
Thanks Manuel for looking into this and providing a fix. Something went wrong with your Pull Request though as it contains 36 commits already on master and can't be merged.

ManuelB:
Hi Christian,
the reason for this is because I only fixed the 3.1.2.Final branch.

I will generate a new pull request that directly uses master.

/Manuel

Issue #126, opened 2015-03-13
When using a lot of entities (400) it can happen that the GeneratedErraiEntityManagerFactory hits the java method lmit


ManuelB:
```
[INFO]    Adding '1' new generated units
[INFO]       Validating units:
[INFO]          [ERROR] Errors in 'application/target/.generated/org/jboss/errai/jpa/client/local/GeneratedErraiEntityManagerFactory.java'
[INFO]             [ERROR] Line 410: The code of method createMetamodel() is exceeding the 65535 bytes limit
[INFO]             See snapshot: /var/folders/bj/5rcf2yl51vx9hhxrxbwdm57r0000gn/T/org.jboss.errai.jpa.client.local.GeneratedErraiEntityManagerFactory910634265386393534.java
[INFO]    [ERROR] Errors in 'application/target/.generated/org/jboss/errai/jpa/client/local/GeneratedErraiEntityManagerFactory.java'
[INFO]       [ERROR] Line 410: The code of method createMetamodel() is exceeding the 65535 bytes limit
[INFO]       See snapshot: /var/folders/bj/5rcf2yl51vx9hhxrxbwdm57r0000gn/T/org.jboss.errai.jpa.client.local.GeneratedErraiEntityManagerFactory8110347497808985426.java
[INFO]    [ERROR] Errors in 'org/jboss/errai/jpa/client/local/ErraiEntityManagerProvider.java'
[INFO]       [ERROR] Line 24: Rebind result 'org.jboss.errai.jpa.client.local.GeneratedErraiEntityManagerFactory' could not be found
[INFO]    Computing all possible rebind results for 'org.uberfire.client.resources.WorkbenchResources'
[INFO]       Rebinding org.uberfire.client.resources.WorkbenchResources
[INFO]          Invoking generator com.google.gwt.resources.rebind.context.InlineClientBundleGenerator
[INFO]             Preparing method CSS
[INFO]                The following problems were detected
[INFO]                   [WARN] Line 27 column 25: encountered "=". Was expecting one of: "+" "-" "," "/" ")" <STRING> <IDENT> <NUMBER> <URL> <PERCENTAGE> <PT> <MM> <CM> <PC> <IN> <PX> <EMS> <EXS> <DEG> <RAD> <GRAD> <MS> <SECOND> <HZ> <KHZ> <DIMEN> <HASH> <UNICODERANGE> <FUNCTION> 
[INFO]    Adding '1' new generated units
[INFO]       Validating units:
[...]

PR #128, opened 2015-03-13
 #126 create own method for every entity type to prevent java function limit of 65536 bytes

ManuelB:
This is for bug #126

### DIFF
--- a/errai-jpa/errai-jpa-client/src/main/java/org/jboss/errai/jpa/rebind/ErraiEntityManagerGenerator.java
+++ b/errai-jpa/errai-jpa-client/src/main/java/org/jboss/errai/jpa/rebind/ErraiEntityManagerGenerator.java
@@ -185,7 +185,7 @@ public class ErraiEntityManagerGenerator extends AbstractAsyncGenerator {
       entityTypes.add(mt.getJavaType());
 
       // first, create a variable for the EntityType
-      String entityTypeVarName = generateErraiEntityType(et, cmm, globalEntityListeners);
+      String entityTypeVarName = generateErraiEntityType(et, cmm, globalEntityListeners, classBuilder);
 
       MethodBodyCallback methodBodyCallback = new JpaMetamodelMethodBodyCallback(classBuilder, et);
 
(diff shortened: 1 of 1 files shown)


csadilek:
Thanks, Manuel. I've looked into this now. I think it's a good start but doesn't fully solve the method byte limit problem yet. You have moved the creation of the anonymous inner classes (entity types) to newly generated private methods. However, the code of these anonymous inner classes didn't count towards the method limit of `createMetaModel` (you can verify this using javap) as inner classes are compiled into separate files. To address this I would suggest to also move the corresponding `addAttribute` calls to your newly generated methods. What do you think?

errai-jpa/errai-jpa-client/src/main/java/org/jboss/errai/jpa/rebind/ErraiEntityManagerGenerator.java
```diff
@@ -241,18 +241,23 @@ public static EntityManagerFactory createHibernateEntityManagerFactory(TreeLogge
...
-  private String generateErraiEntityType(final EntityType<?> et, MethodBlockBuilder<?> pmm, List<MetaClass> globalListeners) {
+  private String generateErraiEntityType(final EntityType<?> et, MethodBlockBuilder<?> pmm, List<MetaClass> globalListeners, ClassStructureBuilder<?> classBuilder) {
     MetaClass met = MetaClassFactory.get(et.getJavaType());
     pmm.append(Stmt.codeComment(
         "**\n" +
         "** EntityType for " + et.getJavaType().getName() + "\n" +
         "**"));
     String entityTypeVarName = entitySnapshotVarName(et.getJavaType());
+    
+    String entityTypeMethodName = "createEntityType"+et.getJavaType().getName().replace('.', '_');
```
csadilek: The name of this new method should prob. be "createEntityType_" + .... for easier readability: createEntityTypeorg_jboss_errai_demo_client_shared_UserComplaint() vs. createEntityType_org_jboss_errai_demo_client_shared_UserComplaint()

ManuelB:
Hi Christian,
go ahead and do all necessary changes. I can just say that my program is compiling. So it seems that my changes influenced the size of the method :-)

Is it possible to release this in the 3.1.2 branch as well in 3.1.3?

/Manuel

csadilek:
Yeah, but it doesn't actually fix the issue as you or someone else will likely run into this having more entities on the classpath. BTW, do you intend to use JPA on the client-side (actually persist some of your entities in the browser's offline storage)? Because if you don't, you do not need this file at all and you're needlessly increasing your JS file size.

ManuelB:
Hi Christian,
actually this is the number one reason why we are choosing Errai. Currently the problem is gone. If it comes again I will fix it again. I would love to spend time enhancing errai but I have to concentrate on business logic.

/Manuel

PR #129, opened 2015-03-17
Remove java 8 incompatible generic signatures

JamesXNelson:
The generics removed in this PR are not supported in current Java 8 Gwt compilation.

If interested, the only other things I needed to do for Java 8 was to update javassist to the latest version.

### DIFF
--- a/errai-cdi/errai-cdi-client/src/main/java/org/jboss/errai/enterprise/client/cdi/InstanceProvider.java
+++ b/errai-cdi/errai-cdi-client/src/main/java/org/jboss/errai/enterprise/client/cdi/InstanceProvider.java
@@ -72,16 +72,16 @@ public class InstanceProvider implements ContextualTypeProvider<Instance> {
     }
 
     @Override
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "rawtypes"})
     public boolean isUnsatisfied() {
-      Collection<IOCBeanDef<?>> beanDefs = IOC.getBeanManager().lookupBeans(type, qualifiers);
+      Collection beanDefs = IOC.getBeanManager().lookupBeans(type, qualifiers);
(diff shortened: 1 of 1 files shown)


csadilek:
Thank you, James!

JamesXNelson:
Cool.  Do you want me to look into upgrading javassist, or is that something better done by a domain expert in the jboss realm?

JamesXNelson:
Using lambdas + CDI = a joy to code :D

Thanks for this awesome project.

csadilek:
Sure, if you want to try upgrading. I was just about to give it a try and run our integration tests: mvn -Pintegration-test clean install (which takes between 1 - 3 hours depending on your hardware).

csadilek:
I guess let me try it on our build server then it won't stop you from doing something else :).

JamesXNelson:
Cool. I'm lookin' forward to it.  Had trouble building drools and jbpm against snapshots, but I'm successfully using java 8 in an uberfire app I'm working on.

I'm pretty newb on the jboss stack, but I tried upgrading to latest Errai, Javassist and Gwt, and the IoC generator was having troubles looking up some classes.  The problem was similar to https://developer.jboss.org/thread/250279 .  I tried the solution (using Impl class) as done [here](https://github.com/droolsjbpm/guvnor/blob/6.2.x/guvnor-asset-mgmt/guvnor-asset-mgmt-client/src/main/java/org/guvnor/asset/management/client/editors/repository/structure/RepositoryStructurePresenter.java#L118), but it was beyond my abilities to fix within the allotted 2 days I gave myself, so I gave up and built the latest release instead.

csadilek:
OK pushed an upgrade to javassist 3.19.0-GA after all tests passed. New snapshots are out too. If you still experience the IoC issue let me know. I can take a look at that.

PR #131, opened 2015-04-10
define guice dependency scope as provided everywhere

porcelli:


csadilek:
Sorry, can't do this as it will break all demos. The server-side message bus has a dependency on guice and therefore needs it at runtime. We need to investigate an alternative solution for your problem.

PR #136, opened 2015-06-28
Fixed generic array handling in GWTParameterizedType.getTypeParameters.

axls:
GWTParameterizedType.getTypeParameters throw IllegalArgumentException
for generic array type parameter.


csadilek:
Thank you!

PR #140, opened 2015-09-05
Work around for ERRAI-872 and fix for websocket re-connect after network failure

lorinmetzger:


mbarkley:
Hi Lorin,

Thanks for the PR! I would like to merge your workaround for ERRAI-872 right away, but I have reservations about the other commit with the websocket change. Would you please explain the reasoning behind resending a successful negotiation message? It seems as though that block is only hit if the websocket is already active. Would you mind explaining to me what scenario this fixes?

If you would like to submit a separate PR for the ERRAI-872 workaround, I'd be happy to merge that right away while we discuss the other changes.

Cheers.

mbarkley:
Lorin,

I just saw ERRAI-873 where you described the issue. Sorry I missed that.

But I am wondering, is there any way we could add a meaningful test for this?

lorinmetzger:
Hi Max,


I'll try to submit a separate PR for ERRAI-872 when I get home later
tonight.  I should kept the two separate.

I'm not sure if we could build a closed loop test case for ERRAI-873.   I
was testing it on a iPad, where I was connected via WIFI, then went to
settings and turned WIFI off, to force it back to a cellular data
connection, and when it tried to reconnect over the cellular network it
would never successfully negotiate a new connection, because the server
would never send the successful negotiation response.   I think you could
do a similar thing on a laptop by having WIFI and Eithernet connected and
switching b/w the two.

You basically need to not so gracefully kill and re-establish your network
connection.

Thanks for working on this!

-Lorin

lorinmetzger:
I opened a new pull request that just has the change to the ClientSyncManager.   I do believe there is still be bug in the Websocket handler, ref ERRAI-873.

mbarkley:
Lorin,

After some thought, I can't come up with a good automated test for the Websocket issue. Since this does seem like a fix we should have in Errai, I would like to merge this. I am reopening the issue, but I am hoping that you could amend the Websocket commit to mention the Jira issue (ERRAI-873) before I merge it.

lorinmetzger:
@mbarkley,

I created a separate pull request for ERRAI-873 #145 

On a separate issue in the same class, I believe that line 88 should be removed from WebSocketNegotiationHandler.java as well.   The only way it could ever be called would be if the activation key doesn't match the session, but in that case a FailedNegotiationMessage would have already been sent back to the client, so it doesn't seem like it makes sense to send a successful response in that case.

There is a comment about this in my original pull request #140 if you get the time to review it, but I didn't commit that change in the new pull request #145 

Thanks!

mbarkley:
Ok. I've merged your other PR for ERRAI-873. Thanks again, Lorin!

PR #144, opened 2015-10-14
ERRAI-872 allows caller to clear syncInProgress flag...

lorinmetzger:
…s transport failed before the sync completed and without the ErrorCallback being called.

### DIFF
--- a/errai-jpa/errai-jpa-datasync/src/main/java/org/jboss/errai/jpa/sync/client/local/ClientSyncManager.java
+++ b/errai-jpa/errai-jpa-datasync/src/main/java/org/jboss/errai/jpa/sync/client/local/ClientSyncManager.java
@@ -249,6 +249,19 @@ public class ClientSyncManager {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lorinmetzger:
Max, I noticed your comment on JIRA about having the ClientSyncManager listen for bus change events.  Let me know if this is something you would like me to work on.

mbarkley:
Thanks, Lorin!

If you're interested making the ClientSyncManager aware of the bus state, we'd definitely appreciate your help.

As a starting point, [here](https://github.com/errai/errai/blob/master/errai-bus/src/main/java/org/jboss/errai/bus/client/api/ClientMessageBus.java#L50-L60) is how you add lifecycle listener to the bus. And here is [the lifecycle interface](https://github.com/errai/errai/blob/master/errai-bus/src/main/java/org/jboss/errai/bus/client/api/BusLifecycleListener.java) which gives a decent overview of the state transitions.

If you want to tackle this and have any questions you can always find me on the Errai IRC channel (#errai on Freenode).

Cheers.

PR #146, opened 2015-11-03
Remove dependency on jboss-as-dist

psiroky:
* `jboss-as-dist` dependency does not seem to be used for testing and it should not be required for runtime either
* it is quite big (more than 100MiB) and it is being transitively dragged into other modules which use `errai-cdi-jboss`


mbarkley:
Thanks, Petr!

PR #148, opened 2015-11-05
[3.2] Remove dependency on jboss-as-dist

psiroky:
Same changes were recently merged into master.


mbarkley:
Thanks again, Petr.

PR #155, opened 2016-01-20
Errai 3.2.2.Final Keycloak 1.7.Final issue - KC roles not obtained

hrstoyanov:
Max,
See if this pull request works for you. Otherwise, just grab the KeycloakAuthenticationService.java from my repo  - the change is small and easily absorbed

### DIFF
--- a/errai-security/errai-security-keycloak/src/main/java/org/jboss/errai/security/keycloak/KeycloakAuthenticationService.java
+++ b/errai-security/errai-security-keycloak/src/main/java/org/jboss/errai/security/keycloak/KeycloakAuthenticationService.java
@@ -16,7 +16,6 @@
  */
 package org.jboss.errai.security.keycloak;
 
-import static org.jboss.errai.security.keycloak.properties.KeycloakPropertyNames.AUDIENCE;
 import static org.jboss.errai.security.keycloak.properties.KeycloakPropertyNames.BIRTHDATE;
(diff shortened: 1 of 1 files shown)


hrstoyanov:
this is re-submit of my previous PR, hopefully easier to merge in 3.2. When can you get it done?

mbarkley:
Hi Hristo,

I'm sorry but I can't merge this as is because the change breaks the errai-security-demo. Following [these steps](https://github.com/errai/errai/blob/3.2/errai-demos/errai-security-demo/README.md#keycloak-integration) (with a Keycloak 1.7 server) I see an NPE on line 230 of the `KeycloakAuthenticationService` after logging in via Keycloak.

Additionally, this change would make the `KeycloakAuthenticationService` use only realm roles (see [here](https://issues.jboss.org/browse/KEYCLOAK-1720)). Ideally we would want to use both realm and client roles. If not that, it should at least be configurable which roles are being used.

Being able to use both client and realm roles would be a enhancement for the framework, so I do hope you are able to address these issues.

Cheers.

hrstoyanov:
I will take a stab at this again. Recently KC also introduced "groups" in addition to roles, so I will try to get all of them (app roles, realm roles, groups)

mbarkley:
Closing this because of the [newer PR](https://github.com/errai/errai/pull/157).

PR #156, opened 2016-01-24
ValueBox<T> implements both HasText and TakesValue<T>.  TakesValue<T>…

thebeeber:
… should take precedence over HasText.  Otherwise, inferWidgetValueType( ValueBox<BigDecimal> w ) will resolve to String, when it should resolve to BigDecimal.  This ends up causing a ClassCastException when binding occurs.

### DIFF
--- a/errai-data-binding/src/main/java/org/jboss/errai/databinding/client/api/Convert.java
+++ b/errai-data-binding/src/main/java/org/jboss/errai/databinding/client/api/Convert.java
@@ -302,10 +302,7 @@ public class Convert {
   public static Class inferWidgetValueType(Widget widget, Class<?> defaultWidgetValueType) {
     Class widgetValueType = null;
 
-    if (widget instanceof HasText) {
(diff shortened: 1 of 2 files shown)


mbarkley:
Hey Josh,

Good catch! I will definitely merge this, but would you be able to add a reproducer test for this issue? That would be very helpful, since I am still working on data-binding changes for our first 4.0 release. 
If you are able, a copy of [this test](https://github.com/errai/errai/blob/master/errai-data-binding/src/test/java/org/jboss/errai/databinding/client/test/DataBindingIntegrationTest.java#L85-L94) using a ValueBox and non-String model property would suffice.

Cheers.

thebeeber:
Let me know if we're good

mbarkley:
Hi Josh,

That's perfect. I have squashed your two commits into [this](https://github.com/errai/errai/commit/d339c6e8c2d1788d94b552cabf3d9515ba93c506) single one on master.

Thanks!

PR #158, opened 2016-02-01
Exclude javassist:javassist which collides with org.javassist:javassist

psiroky:
@csadilek, I've basically replaced the old javassist with new one, so this should be fine. In any case, we can't really have two artifacts with same FQCN on classpath. There is no guarantee which one will be picked up and this leads to very hard to find issues.


mbarkley:
master: 2f4234ef4ee0dbcb6805b01d9c7d8ca540992b7d

Thanks @psiroky !

PR #161, opened 2016-02-26
[DO NOT MERGE] Create non-widget ListComponent and DOM wrapping JsTypes

mbarkley:
These commits introduce changes to the SyncBeanDef interface that require updates to downstream projects (Uberfire, Uberfire Extensions, and Dashbuilder). This PR should not be merged until these downstream changes  are ready to merge as well.


manstis:
Presumably we'll also need to (possibly) change guvnor, kie-wb-common, drools-wb, jbpm-console-ng, jbpm-designer and kie-wb-distributions (depending on their usage of SyncBeanDef)?

mbarkley:
@manstis I think those other repositories are unaffected. This change removes one method and adds another to the SyncBeanDef interface. I couldn't find any usage of the removed method or any implementations of SyncBeanDef in any droolsjbpm repositories.

manstis:
Great stuff - I know of a couple of cases where we get collections of SyncBeanDef but IIRC we then only instantiate the beans... but at least I can keep an eye open should there be problems :-)

csadilek:
master: bf0d507903dfe85844a64c03c21af8d90f87a313 268a454170a167978b78584a13f5856e7ac4a831 7c59d75df0db1487bbeffea3864e442fa3167675

PR #163, opened 2016-03-03
Instance<T> should return the iterator for T implementations.

ederign:


dgutierr:
Related JIRA => https://issues.jboss.org/browse/ERRAI-906

ederign:
I'm o this.

errai-cdi/errai-cdi-client/src/main/java/org/jboss/errai/enterprise/client/cdi/InstanceProvider.java
```diff
@@ -90,7 +90,11 @@ public boolean isAmbiguous() {
 
     @Override
     public Iterator<Object> iterator() {
-      return Collections.emptyList().iterator();
+      Collection beanDefs = IOC.getBeanManager().lookupBeans(type, qualifiers);
+      if(beanDefs==null){
+        return Collections.emptyList().iterator();
+      }
+      return beanDefs.iterator();
```
mbarkley: This line returns an `Iterator<SyncBeanDef>`, but what we want here is an iterator of instances (i.e. `Iterator<T>`).

errai-cdi/errai-cdi-server/src/test/java/org/jboss/errai/cdi/injection/client/test/InstanceInjectionIntegrationTest.java
```diff
@@ -133,4 +133,12 @@ public void testIsAmbiguous() {
     assertFalse(instanceDependentBeanA.isAmbiguous());
   }
 
+  public void testIterator() {
+    final DependentInstanceTestBean testBean
+            = getBeanManager().lookupBean(DependentInstanceTestBean.class).getInstance();
+
+    final Instance<DependentBeanA> injectedBean = testBean.getInjectDependentBeanA();
+
+    assertTrue(injectedBean.iterator().hasNext());
```
mbarkley: Glad to see test coverage, but this test is missing some assertions that would make it much more robust. In particular, it should call `Iterator#next()` and verify that the returned value is a `DependentBeanA`. This would be sufficient to reproduce the issue I described above.

ederign: ok! I'll fix that! @mbarkley  thanks for the review

errai-cdi/errai-cdi-client/src/main/java/org/jboss/errai/enterprise/client/cdi/InstanceProvider.java
ederign: @mbarkley so I should iterate the list and get() all the instances?

mbarkley:
@ederign Thanks for working on this!

As I pointed out in-line, the `Instance#iterator()` method should return an `Iterator<T>`. It may be tempting to make a collection of bean instances and return its iterator but it is very important that the iterator not create instances before the user calls `Iterator#next()`. Creating the instances too early could lead to a memory leak where a bean is created that will never be destroyed and that the user has no reference to.

I'd recommend taking a look at the code in Weld that solves this problem: https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/bean/builtin/InstanceImpl.java#L208-L232

Cheers.

ederign:
That is why I've asked first (memory leak). I'll take a look on the code and fix it!
Cheers.

mbarkley:
@ederign Thanks!

PR #164, opened 2016-03-16
Add missing cdi-client for cdi-server tests

psiroky:
@csadilek, @mbarkley please take a look. This is currently breaking all PR builds as we build Errai from sources.


mbarkley:
Thanks @psiroky !

PR #166, opened 2016-04-02
Upgrade ip-bom to 7.0.0.CR1

psiroky:
* ip-bom import moved into errai-bom so that
   all 3rd party dependencyManagement can happen
   there


mbarkley:
@psiroky Thanks.

PR #167, opened 2016-04-18
Documented template singleton life cycle

christianbauer:


mbarkley:
@christianbauer Thanks!

PR #168, opened 2016-04-22
Upgrade Guava from 19.0.0.jbossorg-1 to 19.0.0.jbossorg-2

psiroky:
To align with Guava version KIE Workbench is now using.


mbarkley:
Thanks @psiroky

PR #175, opened 2016-06-09
PLANNER-579: Reduce WARN messages during startup

mcimbora:
Fixes warnings when starting OptaPlanner WB in Super Dev Mode

WARN  [org.jboss.weld.Event] WELD-000411: Observer method [BackedAnnotatedMethod] public org.jboss.errai.security.server.SecurityAnnotationExtension.addParameterLogger(@Observes ProcessAnnotatedType<Object>) receives events for all annotated types. Consider restricting events using @WithAnnotations or a generic type with bounds.

WARN: can't find jboss-cli.xml. Using default configuration values.

### DIFF
--- a/errai-cdi/jboss/src/main/java/org/jboss/errai/cdi/server/gwt/EmbeddedWildFlyLauncher.java
+++ b/errai-cdi/jboss/src/main/java/org/jboss/errai/cdi/server/gwt/EmbeddedWildFlyLauncher.java
@@ -85,10 +85,16 @@ public class EmbeddedWildFlyLauncher extends ServletContainerLauncher {
   public ServletContainer start(final TreeLogger treeLogger, final int port, final File appRootDir) throws BindException, Exception {
     logger = new StackTreeLogger(treeLogger);
     try {
-      System.setProperty("jboss.http.port", "" + port);
-
       final String jbossHome = JBossUtil.getJBossHome(logger);
       final String[] cmdArgs = JBossUtil.getCommandArguments(logger);
+
+      System.setProperty("jboss.http.port", "" + port);
+
+      File cliConfigFile = new File(jbossHome, JBossUtil.CLI_CONFIGURATION_FILE);
+      if (cliConfigFile.exists()) {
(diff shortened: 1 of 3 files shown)


mbarkley:
Thanks @mcimbora.

mbarkley:
@mcimbora 

The changes to the SecurityAnnotationExtension caused test failures on Jenkins. After adding the `@WithAnnotations( RestictedAccess.class )` the extension does not see classes that implement interfaces annotated with `@RestrictedAccess`. You can see these test failures [here](https://hudson.jboss.org/hudson/job/Errai/1898/) or reproduce them by running `mvn clean install -Pintegration-test` from the `errai-security-client` directory.

I'm not sure how we can remove the warning message for that extension while maintaining functionality, so I have [reverted that part of the commit](https://github.com/errai/errai/commit/1df23d006045e2bd151c0f36b94e5fe6108c454f).

mcimbora:
@mbarkley Thanks for letting me know, I'll have a look into it.

mcimbora:
@mbarkley There are basically 2 options
a) make the classes implement common interface - unacceptable
b) move `@RestrictedAccess` annotation to implementing classes - the contract doesn't specify whether the annotation must be applied on interface level. Still, the change would impact all the clients. Probably not worth the pain.

mbarkley:
@mcimbora I agree with your assessment. The first option isn't viable, and the second option would be a huge pain for both our users. Thanks for investigating.

csadilek:
@mbarkley @mcimbora isn't the problem here just that ```@RestrictedAccess``` is missing ```@Inherited```: https://docs.oracle.com/javase/7/docs/api/java/lang/annotation/Inherited.html

csadilek:
@mbarkley @mcimbora https://issues.jboss.org/browse/CDI-318

csadilek:
@mcimbora @mbarkley actually, this used to work and was documented the way we want it but it does no longer work and the docs have been changed too. So, back to the beginning.

mcimbora:
@csadilek Thanks for the insight.

PR #177, opened 2016-06-13
Fixing properties parser to read properties values ending with '\:' or '\=' correctly

paulovmr:
Bug example: 

Given a properties file:

    key1=value1\:
    key2=value2

The parser would just return one property with key "key1" and value "value1key2=value2" (ignoring the EOL).

### DIFF
--- a/errai-common/src/main/java/org/jboss/errai/common/client/util/Properties.java
+++ b/errai-common/src/main/java/org/jboss/errai/common/client/util/Properties.java
@@ -161,6 +161,12 @@ public class Properties {
               case 'n':
                 state.value.append('\n');
                 break;
+              case '=':
+                state.value.append('=');
+                break;
+              case ':':
+                state.value.append(':');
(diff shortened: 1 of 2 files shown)


paulovmr:
Hi @mbarkley @csadilek, we will have to generate a new tag for errai with this fix, 3.2.4.Final (I guess), and update the [6.4.x pom.xml](https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/pom.xml) with the new version containing this fix.

@mbiarnes This should fix the issue https://issues.jboss.org/browse/RHBPMS-1367 without us having to make any change on Zanata files.

mbarkley:
Thanks @paulovmr. I talked with @mbiarnes and will be doing a 3.2.4.Final release with this fix for the end of next week.

paulovmr:
@mbarkley Nice, thanks!

PR #184, opened 2016-08-05
Update GWT to 2.8.0-rc1 (lots of details below).

mbarkley:
Most of the items below were fixes for tests (which
are now production compiled by default) and dependency
issues (since gwt-user/gwt-dev now bring in transitive
dependencies instead of bundling them in fat jars).

* Upgrade jetty to version in ip-bom
* Fix errai-bus test issues.
* Remove GuardedBy and Nonnull annotations from errai-bus.
* Fix javadoc bug in errai-bus.
* Fix test failures caused by GWT type oracle behaviour change.
(This commit causes the type oracle to hide non-abstract methods
in interfaces:
https://github.com/gwtproject/gwt/commit/75382f1202bf3eaa399d60ebdba42bd7522da3bb)
* Fix AbstractErraiTest so it is usable in errai-ioc-bus-support production compiled tests.
* Fix data-binding tests (have to manually invoke event listeners because dispatchEvent does not work).
* Fixes for errai-cdi-server tests (description below).
* Upgrade Jetty version
* Add -devMode gwt.arg
* Add missing inherited modules
(This was to try and get the tests to run in prod mode,
but there appears to be a compiler bug that prevents this
-- interned value is used before being referenced)
* Use errai.devel.nocache property in Errai GWT generators.
* Fix jaxrs-client bug with Date params.
* Fix errai-ui tests (notes below).
** Fix event dispatching issue (same problem as with data-binding tests).
** Fix NPE when an empty input element fires a change event while bound
to a primitive property in a model.
* Fix errai-navigation native anchor test (same problem as data-binding tests).
* Fix Errai Security tests.
* Fix errai-jpa-client tests.
* Fix gwt-dev version in demos.
* Manage gson version in errai-bom to version required by gwt-dev.


csadilek:
+1

PR #190, opened 2016-09-29
adding support for copying mgmt users and group if provided 

Salaboy:
to the wildfly instance.

@csadilek @mbarkley can you guys let me know if we need a test for this. If so, can you point me out what test is checking the correct behavior for the normal App/Roles files? 

These changes are required to automatically create a users for the wildfly instance that is running in dev mode. The configuration files will be picked up (as the app/roles files) from the src/main/resources of the project starting GWT dev mode.


mbarkley:
@Salaboy I'm happy to merge this as is since it is just re-using the existing logic for loading application users/roles.

PR #193, opened 2016-10-12
Bypasses NPE when the hosting server has no CDI.

smoyer64:
The default Jetty server run in super-dev-mode fails to start with an NPE prior to this change.


mbarkley:
Thanks @smoyer64!

PR #196, opened 2016-10-20
Exclude slf4j-simple from lesscss

manstis:
```lesscss``` brings in dependency ```org.slf4j:slf4j-simple``` which causes upstream build errors (e.g. ```uberfire/uberfire-showcase/uberfire-client-webapp``` where Maven Enforcer Plugins fail the build should a duplicate class implementation be detected). It should be customary for end-user applications to specify logging implementations not transitive libraries.


psiroky:
+1

mbarkley:
Thanks @manstis. Sorry for the trouble!

PR #199, opened 2016-11-02
ObjectMarshaller for null objects causes stack overflow

bbrodt:
See https://issues.jboss.org/browse/ERRAI-990


mbarkley:
Thanks @bbrodt.

PR #206, opened 2016-11-22
ERRAI-1000: User is displayed as ANONYMOUS if his/her role name contains apostrophe

karreiro:
See https://issues.jboss.org/browse/ERRAI-1000

### DIFF
--- a/errai-security/errai-security-server/src/main/java/org/jboss/errai/security/server/servlet/UserHostPageFilter.java
+++ b/errai-security/errai-security-server/src/main/java/org/jboss/errai/security/server/servlet/UserHostPageFilter.java
@@ -48,7 +48,7 @@ import org.jsoup.nodes.Document;
  * application's host page. This is useful in case the login page lives outside
  * the GWT application as it makes instances of {@link User} immediately
  * injectable (without requiring a server round-trip.)
- * 
+ *
  * @author Christian Sadilek <csadilek@redhat.com>
  */
 @WebFilter(filterName = "ErraiUserHostPageFilter", urlPatterns = { "/index.jsp", "/index.html" })
@@ -84,12 +84,11 @@ public class UserHostPageFilter implements Filter {
 
       final User user = authenticationService.getUser();
       final String output;
-      
+
       if (user != null) {
-        final String injectedScript = "<script>var " + 
-                SecurityConstants.ERRAI_SECURITY_CONTEXT_DICTIONARY + "  = {\"" +
-                SecurityConstants.DICTIONARY_USER + "\": '" + 
-                ServerMarshalling.toJSON(user) + "'};" + "</script>";
+        final String injectedScript = "<script>var " +
+                SecurityConstants.ERRAI_SECURITY_CONTEXT_DICTIONARY + "  = " +
+                securityContextJson(user) + "; </script>";
 
         final Document document = Jsoup.parse(wrappedResponse.toString());
         document.head().append(injectedScript);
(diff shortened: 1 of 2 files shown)


manstis:
@karreiro I will leave review to Max however was intrigued by the JIRA. 

[Apparantly](http://www.json.org/) single quotes *should* be permitted in JSON - as ```String``` values should be enclosed in double-quotes. So it's plausible your fix patches the symptom but not the cause and possibly ```ServerMarshalling.toJSON``` may be returning incorrectly formatted JSON in the first place.

karreiro:
@manstis thanks for your comment :-)

Since we have the JSON enclosed in one a single quote String [here](https://github.com/errai/errai/blob/master/errai-security/errai-security-server/src/main/java/org/jboss/errai/security/server/servlet/UserHostPageFilter.java#L91-L92 ), I think that would make sense escaping the ' character in the `UserHostPageFilter` class.

But, let's wait for @mbarkley opinion.

manstis:
@karreiro ah.. perhaps use of the single quote where you linked is the issue; and it should be a double-quote with no need to then escape the single..

mbarkley:
Hi @karreiro,

I think @manstis is correct, and that we really ought to just change the single quotes around JSON value in the filter to double quotes.

karreiro:
Thanks @manstis and @mbarkley for the review.

I think it's done :-)

errai-security/errai-security-server/src/main/java/org/jboss/errai/security/server/servlet/UserHostPageFilter.java
```diff
@@ -98,13 +98,26 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
   }
 
+  String securityContextJson(final User user) {
+    final String userJson = escapedJson(user);
+
+    return "{\"" + SecurityConstants.DICTIONARY_USER + "\": \"" + userJson + "\"}";
+  }
+
+  String escapedJson(final User user) {
+    final LookupTranslator translator = new LookupTranslator(new String[][]{{"\"", "\\\""}});
```
manstis: It looks like you're now also protecting against use of double-quotes in the User's Roles/Groups; however should you not also then be protecting against \ too? Also, isn't a ```String.replaceAll(...)``` sufficient? Does ```LookupTranslator``` really bring anything new to the table!!?

errai-security/errai-security-server/src/test/java/org/jboss/errai/security/server/servlet/UserHostPageFilterTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertTrue;
+
+public class UserHostPageFilterTest {
```
manstis: You explicitly added support for double-quotes in the Users' Role/Groups but fail to include a test for it.

mbarkley:
Hi @karreiro,

Sorry to belabour this whole process, but it now occurs to me that we shouldn't be wrapping the Errai marshalling JSON in any quotes at all. The value returned by `ServerMarhsalling.toJson` is already valid JSON. The only reason we were having any problems is because it was wrapped in quotes, and we should be able to get rid of all the escaping by simply not wrapping the value in quotes.

I amended your commit ([see here](https://github.com/mbarkley/errai/commit/64ca68a8cfd006717413242f4a0295034206f15d)) and it seems to work. I think this is the cleanest solution and would prefer to merge that change.

karreiro:
Hi @mbarkley,

Nice approach, it _really is_ the cleanest. However, I think that we would need more changes to remove the double quotes. I got this error when I tested [the amended commit](https://github.com/mbarkley/errai/commit/64ca68a8cfd006717413242f4a0295034206f15d):
error

The JavaScript object `errai_security_context` is being set, but we got a problem with the IOC container.

mbarkley:
@karreiro Good catch. I forgot to change the client-side storage handler, which was obviously a problem since it expected to parse the user from a String.

I've added another commit to [this branch](https://github.com/mbarkley/errai/tree/ERRAI-1000-master) that updates the storage handler. It required a bit more work since we did not have a marshalling API for demarshalling a JSON object directly, but I think it is worth it to not have to worry about escaping still.

WDYT?

karreiro:
@mbarkley Perfect! :-) Do you prefer to close this PR and open another one with your branch or Can I cherry-pick your commit?

mbarkley:
@karreiro I've pushed the commits directly:
3f346a39192ce646bbdef1a0b9ed1d2b085a5053
4626524bc4066a8b4bab12ea9bb90690b7bf7b83

Issue #207, opened 2016-11-22
Incorrect BindableProxyAgent.findConverter() code for figuring out componentValueType


slavap:
Model field type is Date
``` Widget is HasValue<String> ```
Custom Converter<Date, String> 

findConverter() starts with the following code:
```java
if (component instanceof Widget) {
      componentValueType = Optional.ofNullable(Convert.inferWidgetValueType((Widget) component, propertyType));
    }
```

But my widget.getValue() is null on bind, so Convert.inferWidgetValueType() just returns back propertyType, which is Date, as result exception is raised "Converter widget type, java.lang.String, does not match the required type, java.util.Date"

So Convert.inferWidgetValueType() just could not return defaultWidgetValueType in case widget.getValue() is null, because this method has zero knowledge about widgets used in my application. This code is just wrong:

```java
Object value = ((TakesValue) widget).getValue();
      if (value != null) {
        if (value instanceof WrappedPortable) {
          value = ((WrappedPortable) value).unwrap();
        }
        widgetValueType = value.getClass();
      }
      else if (widget instanceof TextBoxBase) {
        widgetValueType = String.class;
      }
      else if (widget instanceof DateBox || widget instanceof DatePicker) {
        widgetValueType = Date.class;
      }
      else if (widget instanceof CheckBox || widget instanceof ToggleButton) {
        widgetValueType = Boolean.class;
      }
      else if (widget instanceof LongBox) {
        widgetValueType = Long.class;
      }
      else if (widget instanceof DoubleBox) {
        widgetValueType = Double.class;
      }
      else if (widget instanceof IntegerBox) {
        widgetValueType = Integer.class;
      }
      else {
        widgetValueType = defaultWidgetValueType; <<< WRONG!!!
      }
```
What about unknown widget types? How to extend or override this code?

The best way how to resolve this problem IMO is adding custom widget registration to Convert class, so I can register my widgets and their value types.

PR #208, opened 2016-11-23
Convert.inferWidgetValueType() made configurable.

slavap:
Fixes https://github.com/errai/errai/issues/207
The simplest possible fix, with it I can register my custom widgets like this:
```java
Convert.widgetValueTypes.put(MyCustomWidget1.class, String.class);
Convert.widgetValueTypes.put(MyCustomWidget2.class, String.class);
```


mbarkley:
Hi @slavap,

I have an alternative fix that won't require extra code on your part. I think the root cause of this bug is that Errai should infer the component type of your given provider in this scenario.

Here is a PR I have created with my fix: https://github.com/errai/errai/pull/209

Could you please verify that this resolves your problem? If so I would prefer this approach as it will make this "just work".

Cheers.

slavap:
Yes, your fix works fine. Thank you!

PR #209, opened 2016-11-23
Fix default inferred Widget type when user-provided converter is pres…

mbarkley:
…ent.


slavap:
@mbarkley  Yes, your fix works fine. Thank you!

PR #221, opened 2017-01-06
Upgrade IP BOM to 7.0.0.CR6

psiroky:


mbarkley:
@psiroky I ran the integration tests locally and the change to the javax.el artifact breaks the build. The artifact you've replaced it with does not contain an ExpressionFactoryImpl class so it causes ClassNotFound exceptions when running the errai-cdi-server tests.

psiroky:
@mbarkley that's a bummer :( I believe we should generally stick to one sort of spec files. We chose the JBoss ones some time ago in KIE, so that's why did this here. I am not sure if the spec jar should contain impl class `ExpressionFactoryImpl`, but I'll take a look at this.

psiroky:
Oh, it seems the `org.glassfish:javax.el` is actually not a Spec jar, but rather implementation jar! That's pretty confusing :) Anyway, I'll add the jar back since there does not seem to be any other option (WildFly seems to use the glassfish impl as well).

psiroky:
@mbarkley I've reverted the javax.el changes, thanks for spotting this!

psiroky:
@mbarkley please could please give a second look and let me know if the changes are OK?

mbarkley:
Jenkins, please retest this.

psiroky:
I did ran the integ. tests locally and I am getting test failure in  errai-jaxrs-client. However, IIRC this one was failing for me even before, so it should not be related to these changes.

mbarkley:
@psiroky Hi Petr, I am just using this PR to test that the PR build works now. I believe I've fixed the tests that were failing only on Jenkins.

mbarkley:
Jenkins, please retest this.

mbarkley:
(All the tests passed in the last build but the timeout value was too low so it failed while the demos were still building.)

mbarkley:
Jenkins, please retest this.

PR #223, opened 2017-01-11
ERRAI-1017: Replace history token on navigation initialization.

axls:
Replace history token instead of push on navigation initialization.
This one fixes duplicated history entries after loading gwt host-page.


mbarkley:
ok to test

mbarkley:
Thanks @axls!

axls:
@mbarkley Thank you.
Is there a chance to see this commit in 3.2 branch?

mbarkley:
@axls Typically when we plan for a maintenance release, we will go through issues fixed on master and back-port fixes for critical bugs or fixes that are low risk (like this one). We do not currently have another 3.2 release scheduled, but if we do this will be included.

PR #224, opened 2017-01-11
Adding NavigationControl support to @PageShowing

BenDol:
A continuation of #215
This gives an opportunity for "gatekeeping" pages asynchronously.

```java
@PageShowing
void onShowing(HistoryToken historyToken, NavigationControl control) {
    myService.call(hasAccess -> {
        if(hasAccess) {
            // We have access, proceed with navigation.
            control.proceed();
        } else {
            // Interrupt navigation and redirect.
            // can perform any other tasks before redirecting.
            control.redirect(RedirectPage.class);
        }
    }).hasAccess(modelId, userId);
}
```

Tested and confirmed to work in @PageHiding methods too.
Includes tests for each feature.

### DIFF
--- a/errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/Navigation.java
+++ b/errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/Navigation.java
@@ -135,11 +135,6 @@ public class Navigation {
   private boolean locked = false;
 
   /**
-   * Indicates that a navigation request is in hiding phase.
-   */
-  private boolean hiding = false;
-
-  /**
    * Queued navigation requests which could not handled immediately.
    */
   private final Queue<Request> queuedRequests = new LinkedList<>();
@@ -451,7 +446,6 @@ public class Navigation {
     final NavigationControl control = new NavigationControl(Navigation.this, new Runnable() {
       @Override
       public void run() {
-        hiding = false;
         final Access<C> accessEvent = new AccessImpl<>();
 
         accessEvent.fireAsync(component, new LifecycleCallback() {
@@ -490,8 +484,7 @@ public class Navigation {
                 locked = true;
                 hideCurrentPage();
                 request.pageNode.pageShowing(component, request.state, showControl);
-              }
-              catch (Exception ex) {
+              } catch (Exception ex) {
                 locked = false;
                 throw ex;
               }
@@ -501,17 +494,16 @@ public class Navigation {
           }
         });
       }
+    }, new Runnable() {
+      @Override
+      public void run() {
+        hideCurrentPage();
+        setCurrentPage(null);
+      }
     });
 
-    if (!hiding && currentPage != null && currentWidget != null && currentComponent != null && currentWidget.asWidget() == navigatingContainer.getWidget()) {
-      hiding = true;
-      try {
-        currentPage.pageHiding(Factory.maybeUnwrapProxy(currentComponent), control);
-      }
-      catch (Exception ex) {
-        hiding = false;
-        throw ex;
-      }
+    if (currentPage != null && currentWidget != null && currentComponent != null && currentWidget.asWidget() == navigatingContainer.getWidget()) {
+      currentPage.pageHiding(Factory.maybeUnwrapProxy(currentComponent), control);
     } else {
       control.proceed();
     }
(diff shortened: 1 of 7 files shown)


mbarkley:
ok to test

errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/Navigation.java
```diff
@@ -444,49 +448,71 @@ public void onAttachOrDetach(final AttachEvent event) {
...
           }
         });
       }
     });
 
-    if (currentPage != null && currentWidget != null && currentComponent != null && currentWidget.asWidget() == navigatingContainer.getWidget()) {
-      currentPage.pageHiding(Factory.maybeUnwrapProxy(currentComponent), control);
-    }
-    else {
+    if (!hiding && currentPage != null && currentWidget != null && currentComponent != null && currentWidget.asWidget() == navigatingContainer.getWidget()) {
```
mbarkley: I'm not sure I understand the purpose of this `hiding` in this condition. If I am interpreting this change correctly, I think this will introduce problems if you try to navigate away from a page with a `@PageHiding` method having a `NavigationControl`.

For example, say I have a page with a form on it and the page uses `@PageHiding` to block navigation away from the page while the form has not been submitted. With this change if a user tries to navigate a way, the first time it will be blocked by `@PageHiding`, but the second time the `@PageHiding` method won't be called and navigation will just proceed.

errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageShowing.java
```diff
@@ -41,6 +43,14 @@
  * not all history token key names are known at compile time, so
  * {@code @PageState} fields can't be declared to accept their values.
  * <p>
+ * The target method is permitted an optional parameter of type {@link NavigationControl}. If the
+ * parameter is present, the page navigation will not be carried out until
+ * {@link NavigationControl#proceed()} is invoked. This is useful for interrupting page navigation
```
mbarkley: The example here makes doesn't fit as well for `@PageShowing` as for `@PageHiding`. It might be better as, "This is useful for redirecting navigation before a page has been displayed, based on asynchronous logic (for example, to redirect to another page based on the result of an Errai RPC)."

errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/Navigation.java
BenDol: The purpose is when calling `redirect` in `@PageHiding` this causes infinite recursion if we don't ensure the hiding phase isn't executed multiple times. The `hiding` var will always be set to `false` when `proceed` is invoked so there shouldn't be able issues.

As for your use case you are right that would be a problem, any suggestions on the best way to handle this with the use case presented above too?

errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageShowing.java
BenDol: Yeah copy pasted from my `@PageAuthorize` annotation :smile:

errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/Navigation.java
mbarkley: Ah, I think I understand now. Without this `hiding` boolean, calling redirect in a `@PageHiding` method causes a full navigation, meaning another call to `@PageHiding`.

Here's what I recommend: get rid of `hiding` and pass in an `interrupt` Runnable to the contstructor of `control` with a run method as follows.

    public void run() {
      hideCurrentPage();
      setCurrentPage(null);
    }

This way, when you call `Navigation.goTo` in a `redirect`, the `@PageHiding` method will not be called again as there will be no current page (and you will have already hidden it). But if `redirect` is never called, a `@PageHiding` method should still be called every time.

BenDol: Great that sounds like a better solution. Will update that now.

mbarkley:
Merged: 52ad196b70814822dd87cb81a107723939df06a8

Thanks @BenDol!

PR #229, opened 2017-01-27
[ERRAI-1022] Update cdi-demo-tagcloud so that it works on jetty9-server

treblereel:
JIRA: https://issues.jboss.org/browse/ERRAI-1022


mbarkley:
Thanks @treblereel.

PR #235, opened 2017-02-07
ERRAI-1019: +Fix MarshallerModelStrategy (details below)

csadilek:
- No longer stop searching after first portable class was
found, but instead collect all matches (a portable class
could be found in a project that is not a webapp e.g.
errai-common when executed from errai root)

- Avoid writing ServerMarshallerFactory.class  files to
target dirs of projects which are no webapps


mbarkley:
Jenkins, retest this.

PR #238, opened 2017-02-15
Exclude el-api and interceptor-api from cdi-api

psiroky:
We should generally be using the JBoss Spec Jars. These excluded
deps does not seem to be needed, so there is no replacement
for those at this point.

@mbarkley WDYT about this? The other option would be to make it `<scope>provided</scope>`, as I guess we could assume the CDI will be provided by the application server?


mbarkley:
@psiroky Setting the cdi-api jar to provided scope would be a pain since it needs to be on the classpath of a webapp during GWT compilation.

I think this change is fine, but I want to take a closer look later before I merge this.

PR #239, opened 2017-02-21
Upgrade IP BOM from 7.0.0.CR6 to 7.0.0.CR7

psiroky:


mbarkley:
Jenkins, retest this

psiroky:
The build is failing because the CR7 did not get synced into maven central (which is a bug).

@mbarkley I am wondering if it would make sense to add the jboss repo into the errai-bom? (I know this is not ideal, but we use the same in other projects just for cases like this one).

mbarkley:
@psiroky We reference the jboss repo from errai-parent so I don't see any harm in adding it to errai-bom as well. Perhaps you could add that change as an amendment to this PR?

psiroky:
@mbarkley yes, will do in a sec

PR #241, opened 2017-02-28
Fix native stacktraces (chrome only). Make more robust to future chan…

mbarkley:
…ges.


mbarkley:
Jenkins, retest this.

PR #247, opened 2017-04-06
Fixed List component destroyer initialization with Dependent scope

nmirasch:
Changed the condition for the List Component destroyer initialization when its working with elements with Dependent scope


mbarkley:
ok to test

mbarkley:
Thanks @nmirasch!

PR #248, opened 2017-04-19
Default translation (locale == null) should be returned when there is no translation to a key in the active locale and the errai.i18n.default_per_key is set to true.

paulovmr:


paulovmr:
@mbarkley Hi Max, see what you think. This way the default translation would be used every time there is no translation to a key in the active locale.

paulovmr:
@mbarkley Hi Max, I made the modifications that we talked about on IRC, if you could take a look. Thanks.

mbarkley:
ok to test

mbarkley:
I believe the build is stalled because of Nexus outages so I'm merging this now.

PR #250, opened 2017-04-27
Allow redefinition of alternative port as it is already possible via …

ngs-mtech:
Allow reading system property defined to change httpControllerPath based on System Property 
jboss.management.http.port


mbarkley:
ok to test

mbarkley:
Hi @ngs-mtech,

This seems like a good change that I would like to merge, but we prefer to accept PRs to our master branch and backport where necessary. At present, we do not have any more 3.2.x releases planned.

Issue #258, opened 2017-05-10
Getting started broken on errai-ui: Unable to find 'org/jboss/errai/databinding/DataBinding.gwt.xml' on your classpath


ge0ffrey:
I only need to use errai-ui for now, so I followed the instructions on this page:
  http://erraiframework.org/getting-started/index.html#ErraiUIPage
and added the dependency and inherited module, which causes:

```
[INFO] --- gwt-maven-plugin:1.0-rc-6:compile (default-compile) @ optashift-employee-rostering-gwtui ---
[INFO] Loading inherited module 'org.optaplanner.openshift.employeerostering.gwtui.WorkerRosteringWebapp'
[INFO]    Loading inherited module 'org.jboss.errai.ui.UI'
[INFO]       Loading inherited module 'org.jboss.errai.databinding.DataBinding'
[INFO]          [ERROR] Unable to find 'org/jboss/errai/databinding/DataBinding.gwt.xml' on your classpath; could be a typo, or maybe you forgot to include a classpath entry for source?
```

ge0ffrey:
**Workaround**: add this dependency:

```
    <dependency>
      <groupId>org.jboss.errai</groupId>
      <artifactId>errai-data-binding</artifactId>
    </dependency>
```

Looks like errai-ui is missing a transitive non-optional dependency on errai-data-binding.

ge0ffrey:
I've submitted a PR for this: https://github.com/errai/errai/pull/259

Also, looks like that all usages of `<scope>provided</scope>` in dependency elements should be replaced by `<optional>true</optional>`.

mbarkley:
Hi @ge0ffrey,

Closing this issue as I've copied it to [JIRA](https://issues.jboss.org/browse/ERRAI-1045).

PR #259, opened 2017-05-10
errai-data-binding is a non-optional dependency for errai-ui

ge0ffrey:
See https://github.com/errai/errai/issues/258


mbarkley:
ok to test

PR #262, opened 2017-05-11
ERRAI-1043: Getting started broken: Do not depend on xercesImpl 2.11.…

mbarkley:
…0.SP4 because that doesn't exist in Maven Central

### DIFF
--- a/errai-bom/pom.xml
+++ b/errai-bom/pom.xml
@@ -64,6 +64,9 @@
     <version.javax.validation>1.0.0.GA</version.javax.validation>
     <version.org.glassfish.javax.el>3.0.0</version.org.glassfish.javax.el>
     <version.org.lesscss>1.7.0.1.1</version.org.lesscss>
+    <!-- The version of xercesImpl in the ip-bom is not published on Maven Central.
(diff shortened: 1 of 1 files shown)


mbarkley:
@psiroky I would like to get your feedback before I merge this.

This version override fixes the issue @ge0ffrey discovered, where the ip-bom overrides the version of `xerces:xercesImpl` to `2.11.0.SP4` which is not in Maven Central. This breaks the Errai getting started experience as user apps won't build without first adding the JBoss Nexus repo to their POMs.

Of the solutions originally proposed by @ge0ffrey, I believe this to be the best one because:

* This is only an issue for Errai community users, so I don't see the need to update the ip-bom.
* AFAIK, nothing in the platform uses errai-bom for version management so this won't affect our team.
* The GWT team is unlikely to do anything about this, because even if the newer version of xercesImpl contains security fixes it is only used in gwt-dev which is never deployed as part of an app.
* Presumably this is actually used by gwt-dev somewhere so excluding it seems like it could just cause other problems.

Please let me know if you see any flaws in my reasoning.

psiroky:
Looks good to me.

PR #263, opened 2017-05-12
ERRAI-1035: @IgnoreBindings annotation for BindableProxyGenerator to skip annotated methods

saso5:


mbarkley:
ok to test

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/TestModelWithIgnoreAnnotation.java
```diff
@@ -0,0 +1,44 @@
+package org.jboss.errai.databinding.client;
```
mbarkley: Please add a license header to this file. (You can copy it from any of the other Java source files in this project, but remember to change the year.)

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/BindableProxyGenerator.java
```diff
@@ -238,6 +239,10 @@ private void generateGetter(final ClassStructureBuilder<?> classBuilder, final S
       final BlockBuilder<?> getMethod) {
 
     final MetaMethod getterMethod = bindable.getBeanDescriptor().getReadMethodForProperty(property);
+    if(isToBeIgnored(getterMethod)){
```
mbarkley: I think it would be better to perform this check at the call-site of `generateGetter`. The issue with the logic currently is that it is now possible to have a model at run-time that has a generated setter but no generated getter. You can see in `generateSetter` that the previous logic ensured that a setter was only generated if a getter method existed.

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/BindableProxyGenerator.java
```diff
@@ -253,13 +258,22 @@ private void generateGetter(final ClassStructureBuilder<?> classBuilder, final S
...
+
   /**
    * Generates a setter method for the provided property plus the corresponding code for the
    * implementation of {@link HasProperties#set(String, Object)}.
    */
   private void generateSetter(final ClassStructureBuilder<?> classBuilder, final String property, final BlockBuilder<?> setMethod) {
     final MetaMethod getterMethod = bindable.getBeanDescriptor().getReadMethodForProperty(property);
     final MetaMethod setterMethod = bindable.getBeanDescriptor().getWriteMethodForProperty(property);
+    
+    if(isToBeIgnored(setterMethod)){
```
mbarkley: Please see comment in `generateGetter`. If that check is moved it would make sense to move this one as well.

errai-data-binding/src/main/java/org/jboss/errai/databinding/client/api/IgnoreBinding.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.jboss.errai.databinding.rebind.BindableProxyGenerator;
+
+/**
+ * Indicates that method or field should be ignored by the {@link BindableProxyGenerator}
```
mbarkley: This javadoc is too implementation specific. Most of our users don't know what the `BindableProxyGenerator` is (and that's probably a good thing!).

From my reading of the code, this annotation:
* When put on a bean accessor (getter or setter) causes Errai to ignore that accessor. If a properties accessors are ignored, it will not be considered a binable property.
* When put on a non-accessor, causes Errai to not proxy that method. Normally Errai proxies non-accessor methods so that we can check for model updates that happen outside of getter and setter calls.

Could you please update the javadoc to describe this behaviour?

errai-data-binding/src/main/java/org/jboss/errai/databinding/client/api/IgnoreBinding.java
```diff
@@ -0,0 +1,19 @@
+package org.jboss.errai.databinding.client.api;
```
mbarkley: This file also needs a license header.

saso5:
@mbarkley thank you for your comments. Will try to implement them as soon as possible.

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/TestModelWithIgnoreAnnotation.java
```diff
@@ -0,0 +1,85 @@
...
+		return null;
+	}
+	
+	@IgnoreBinding
+	public void set(String property, Object value) {
+		return;
+	}
+	
+	@IgnoreBinding
+	public Integer getId() {
```
mbarkley: Please see comment on `generateNonAccessorMethods`. We should write a test for this method that verifies that `getId` dispatches to the actual model and not the proxy. Since `setId` does not have the annotation, it will be properly proxied, so you could call `setId` with some value and verify that `getId` returns the same value.

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/test/HasPropertiesIntegrationTest.java
```diff
@@ -112,5 +113,23 @@ public void testSetThrowsNonExistingPropertyException() {
...
+    final Map<String, PropertyType> properties = model.getBeanProperties();
+    final Set<String> actualProperties = properties.keySet();
+    final Set<String> expectedProperties = new HashSet<String>(Arrays.asList("properties", "name", "age"));
+    
+    assertEquals(expectedProperties, actualProperties);
+    
+    try {
+    	model.set("name", "Test");
+    	fail("Setter should not be present");
+    } catch(NonExistingPropertyException ex) { }
```
mbarkley: Minor nit-pick, but please add spaces around the parentheses.

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/TestModelWithIgnoreAnnotation.java
saso5: I've been trying to solve this issue but as far as I can tell this cannot be done without changing the bindable proxy generator. Because calling "ignored methods" on a proxy of a model does not reach the model, because the model is stored in "agent.target". 

So the generator should always create the methods, just not execute the "agent.updateWidgetsAndFireEvent".

But if the methods have to stay, I'm back at square one with my existing problem. Having two methods with the same signature (get("key"), set("key", "value")).

I could change the generators and leave the feature if you think it could be useful, but to fix my problem I would have to add a condition not to generate the BindableProxy specific methods.

mbarkley: If you add `@IgnoreBinding` on a non-property method then I think it's fine for the data-binding generator to just ignore that method and generate nothing for it. That will fix your issue, since every method name that conflicts with the `BindableProxy` will not be a getter or setter.

But if you ignore an accessor, it should still be usable, just not as a property of the model (so it should still be proxied). If we could make that change, then I would be satisfied with this feature.

Another option, to enhance this, is we could add a property to the annotation to specify whether method should be ignored as an accessor, or fully ignored. I don't think this is necessary, but it's an option.

saso5: I've assumed that method get("key") would be a accessor method, but I've tested the generator and it's not; is grabbed by non-property method generator.

But still, I don't see how to propagate the changes of a non-property method to the source. Here it the test I've been working on.  Because the method that is ignored is not proxied try's to access the property 'age' that is null, because the model value is in 'agetn.target'. 

image

image

tiagobento:
Closing due to #326.

PR #265, opened 2017-06-06
Make sure the bus content is valid before processing

psiroky:


errai-bus/src/main/java/org/jboss/errai/bus/client/framework/transports/HttpPollingHandler.java
```diff
@@ -340,6 +342,18 @@ public boolean canWait() {
   public void handleProtocolExtension(final Message message) {
   }
 
+  private static void validateContentType(final Response response) {
+    int statusCode = response.getStatusCode();
+    // in case the response is OK (200), then the content type always has to be 'application/json'
+    if (statusCode == 200) {
```
psiroky: @mbarkley I am not sure about this. In case the statusCode is .e.g `404`, then content will of course not be `application/json`, but we still want to continue as the client side will try to recover from this (as it may just be a temporary network issue). We definitely only need to check some status code, but is 200 the only one?

psiroky:
Please hold on with this one. I just found out that it solves the issue in one showcase, but fails in some case. Will need to investigate that.

psiroky:
So in some cases (e.g. drools-wb-webapp) the client gets a `403` response saying `Forbidden`. However, in other cases (e.g. kie-wb-webapp) client gets a `200` with a login page HTML content. This is likely related to some configuration.

mbarkley:
@psiroky Looking at this with fresh eyes, we probably only want to make two changes:

1. Throw the new exception you've made when a response with status 200 is malformed.
2. Make sure we aren't trying to parse a payload when the response status isn't 200.

For (2), here are some places we'll need to fix:

* [This](https://github.com/errai/errai/blob/master/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/transports/HttpPollingHandler.java#L416-L417) should be removed (including `break`) so that it falls through to the `onError` call.
* [This](https://github.com/errai/errai/blob/master/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/transports/HttpPollingHandler.java#L797) line should have a conditional guard to assure that it only happens for valid response with status 200.

With these changes, the workbench should be able to show the proper prompt to users by:

* Catching the `InvalidBusContentException` in the default error handler as previously discussed.
* Registering a [lifecycle listener](https://github.com/errai/errai/blob/master/errai-bus/src/main/java/org/jboss/errai/bus/client/api/ClientMessageBus.java#L60) for when the bus goes offline.

psiroky:
@mbarkley could you please take another look?

mbarkley:
@psiroky The change looks good to me. That being said any time there are changes to the message bus I like to do some manual testing before merging. I will try and find time for that this afternoon.

mbarkley:
@psiroky Well it looks like this change breaks bus reconnection after a session expires (in the case that the bus isn't secured). I'll look into this before I sign off today.

psiroky:
@mbarkley thanks, I will check the failing tests.

mbarkley:
@psiroky Found the issue. When Errai's message bus servlet receives a request from a client with an expired session, it sends back a bus payload but the status is 401. This payload has a SessionExpired command that needs to be processed for the bus to attempt reconnection.

IMO there is no value to responding with 401 in this situation since the response is returning a valid payload so [here](https://github.com/mbarkley/errai/commit/ee62333830123f522e09b6b271badc2786b41285) is my fix.

Maybe you could try this out and, if everything works on your end, amend the PR? I will also do some more testing on my end, but probably will have to continue on Monday.

psiroky:
@mbarkley is this still something that could be merged? Or do we need different approach?

I'll need to test this again with the WB webapps, but before that I would like to make sure this is the way to go.

mbarkley:
@psiroky Yes I think this is the right approach. If you try it out on the workbench without issues, then I am good to merge this.

psiroky:
@mbarkley thanks, I'll test with one of the WB apps and let you know.

psiroky:
@mbarkley I believe this should be ready. I am able to catch the newly introduced exception and show a modal based on that, requesting users to log in again.

mbarkley:
@psiroky Merged. Do we need this back-ported to the 4.0.x branch for 7.3.x?

psiroky:
@mbarkley yes. I'll create new PR for that.

PR #266, opened 2017-06-08
JBPM-5861: Added JsInterop support for constructing Blob types and for the Window.atob function.

romartin:
Hey @mbarkley !

This commit includes a few `jsinterop` types and methods, I need those to achieve some features from [this ticket](https://issues.jboss.org/browse/JBPM-5861).

Thanks!
Roger


romartin:
ok to test

mbarkley:
Ok to test

PR #269, opened 2017-07-14
Fix for marshall() and demarshall() generated methods. In my project …

slavap:
…resulting js code size went down from 433kb to 380kb, and demarshall() performance few times better now.

Fixes both:
https://issues.jboss.org/browse/ERRAI-1055
https://issues.jboss.org/browse/ERRAI-892

### DIFF
--- a/errai-marshalling/src/main/java/org/jboss/errai/marshalling/rebind/api/impl/defaultjava/DefaultJavaMappingStrategy.java
+++ b/errai-marshalling/src/main/java/org/jboss/errai/marshalling/rebind/api/impl/defaultjava/DefaultJavaMappingStrategy.java
@@ -23,6 +23,7 @@ import static org.jboss.errai.codegen.util.Stmt.loadVariable;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import javax.enterprise.util.TypeLiteral;
 
 import org.jboss.errai.codegen.Cast;
 import org.jboss.errai.codegen.InnerClass;
@@ -270,19 +271,7 @@ public class DefaultJavaMappingStrategy implements MappingStrategy {
               loadVariable("objId"), loadVariable("entity")));
         }
         
-        CaseBlockBuilder switchBlock;
-        builder.append(StatementBuilder.create()
-            .loadVariable("obj").invoke("keySet")
-            .foreach("key", String.class)
-            .append(Stmt.if_(Bool.or(loadVariable("key").invoke("equals", SerializationParts.ENCODED_TYPE), 
-                                     loadVariable("key").invoke("equals", SerializationParts.OBJECT_ID)))
-                        .append(Stmt.continue_()).finish())
-            // objVal assignment in each switch's case clause leads to excessively large resulting js size!
-            .append(Stmt.declareVariable("objVal", EJValue.class, extractJSONObjectProperty(loadVariable("key"), 
-                    EJObject.class, "getIfNotNull")))
-            .append(Stmt.if_(Bool.isNull(loadVariable("objVal"))).append(Stmt.continue_()).finish())
-            .append(switchBlock = Stmt.switch_(loadVariable("key")))
-            .finish());
+        CaseBlockBuilder switchBlock = null;
 
         /**
          *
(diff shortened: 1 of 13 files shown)


mbarkley:
ok to test

mbarkley:
@slavap Also there seem to be test failures now. Unfortunately the PR build isn't setup to show codegen stack traces.

You should try running this locally to see if you can reproduce the issues:

    mvn clean install -Pintegration-test -Derrai.codegen.details=true

mbarkley:
jenkins retest this

slavap:
@mbarkley 

> But more importantly, I have concerns about changing from append to concatenation.
> Does this change impact runtime performance when marshalling/demarshalling? 
> Are we sure that the code generated from concatenation is as efficient as using the StringBuilder?

StringBuilder is not native to javascript, so could not be optimized by js vm.
From java7 string concatenations are automatically processed by JVM as StringBuilder under the hood.
So IMO all this append() stuff makes sense for java6 only.
Generated js code is very clean and much smaller after my fix.

> In what environment did you test the code-size difference? More specifically: How big was the project? How many portable types? Did you do a full compile with optimisations and obfuscation?

I've tested in my current project. Full optimization and obfuscation of course.
About 270 portable classes, some classes with more than 100 properties.

> I can try and find time this week to help do some profiling.
Just look at resulting js code (full optimization, pretty style) and it will be clear how better newly generated js code is.

slavap:
@mbarkley I've added changes according to your request. Unfortunately I still see test failures :-( Tried to run: mvn clean install -Pintegration-test -Derrai.codegen.details=true 
on my PC without success, it fails in testPointToPointMessageInClusterAfterClientChangesNodes, but I cannot figure out what/where the problem is :-( I need help with that.
mvn clean install - works just fine, no errors.

errai-marshalling/src/main/java/org/jboss/errai/marshalling/rebind/api/impl/defaultjava/DefaultJavaMappingStrategy.java
```diff
@@ -262,11 +269,25 @@ else if (instantiationMapping instanceof FactoryMapping) {
           builder._(loadVariable("a1").invoke("recordObject",
               loadVariable("objId"), loadVariable("entity")));
         }
+        
+        CaseBlockBuilder switchBlock;
+        builder.append(StatementBuilder.create()
+            .loadVariable("obj").invoke("keySet")
+            .foreach("key", String.class)
```
mbarkley: @slavap The error stacktrace points to this line. For some reason Errai Codegen thinks the actual type of the loop variable is `Object`.

I would suggest working around this by chainging this line to:
    .foreach("key", Object.class)

Then use `Stmt.cast` below where necessary.

mbarkley:
@slavap 

I have added an inline comment in `DefaultJavaMappingStrategy` about the cause of the test errors.

I see what you are saying about StringBuilder in the JVM, but I am less certain about what code gets generated by GWT using StringBuilder versus concatenation (as clearly the code must be different, or else it wouldn't have been larger with StringBuilder). I will try to do some tests of my own to empirically verify that there is no runtime performance regression.

slavap:
@mbarkley I will check how using cast will affect generated js code. Not sure it's a good idea. Would you like to see the example of generated js code? Actually performance after my fix even visually noticeable, first I've thought that marshaling started passing empty :-)

slavap:
@mbarkley Replaced: for (String key : obj.keySet()) { ... }
by: final Set<String> keys = obj.keySet();  for (String key : keys) { ... }
That way no cast is needed and integration tests should pass.

slavap:
@mbarkley At last "All checks passed"! 
By the way, on my PC it is not that good :-( 
For example Errai::Bus::JSR-356::WebSocket::Weld ............... FAILURE with: 

Running org.jboss.errai.bus.server.websocket.test.jsr356.cdi.filter.FilterTest
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.232 sec <<< FAILURE! - in org.jboss.errai.bus.server.websocket.test.jsr356.cdi.filter.FilterTest
org.jboss.errai.bus.server.websocket.test.jsr356.cdi.filter.FilterTest  Time elapsed: 1.232 sec  <<< ERROR!
org.jboss.arquillian.container.spi.client.container.DeploymentException: Cannot deploy: filtertest.war
Caused by: java.io.IOException: Invalid response

God knows what that means :-(
And integration tests are painfully slow, even on decent 4 cores PC with SSD.

Final results of my fix - resulting js down from 433kb to 362kb, i.e. > 16% reduction.
Next step is data bindings optimization :-)

PR #272, opened 2017-07-27
Fix Jenkins CI build

tiagobento:
- Adding deployment skip configuration to `errai-demos/errai-security-demo` module.

- Adding `<distributionManagement />` configuration to `errai-forge-addon` module.

- Removing unnecessary `<distributionManagement />` configuration in `errai-client-local-class-hider` because it's inherited from `org.jboss:jboss-parent`. _(not necessary to fix Jenkins CI build)_

- Removing unnecessary dependency `junit:junit` in `errai-client-local-class-hider` because it's inherited from `errai-parent`. _(not necessary to fix Jenkins CI build)_

### DIFF
--- a/errai-demos/errai-security-demo/pom.xml
+++ b/errai-demos/errai-security-demo/pom.xml
@@ -62,6 +62,7 @@
     <javaee.version>3.0.1.Final</javaee.version>
     <errai.dev.context>${project.artifactId}</errai.dev.context>
     <maven.gwt.plugin.version>2.8.0</maven.gwt.plugin.version>
+    <maven.deploy.plugin.version>2.7</maven.deploy.plugin.version>
   </properties>
 
   <repositories>
@@ -308,6 +309,14 @@
           <soyc>false</soyc>
           <hostedWebapp>src/main/webapp</hostedWebapp>
           <extraJvmArgs>-Xmx1G -XX:CompileThreshold=7000 -Derrai.jboss.home=${errai.jboss.home} -Derrai.dev.context=${errai.dev.context}</extraJvmArgs>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>${maven.deploy.plugin.version}</version>
+        <configuration>
+          <skip>true</skip>
         </configuration>
       </plugin>
     </plugins>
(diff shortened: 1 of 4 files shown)


errai-demos/errai-security-demo/pom.xml
```diff
@@ -310,6 +311,14 @@
           <extraJvmArgs>-Xmx1G -XX:CompileThreshold=7000 -Derrai.jboss.home=${errai.jboss.home} -Derrai.dev.context=${errai.dev.context}</extraJvmArgs>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>${maven.deploy.plugin.version}</version>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
```
psiroky: I am not 100% sure this will pass productized build. IIRC skipping deployment fails the build there. @ryanzhang or @AntStephenson could you please clarify?

tiagobento: @psiroky The other demos do skip the deployment, so I figured I should replicate the behavior. See https://github.com/errai/errai/blob/master/errai-jaxrs/errai-jaxrs-client/pom.xml for instance

psiroky: Ah, I wasn't aware of that. If we already do it in different modules, then this should be fine.

tiagobento: @psiroky Let's hope for it :)

AntStephenson: Sorry for the late reply, I didn't spot this earlier for some reason.. @psiroky AFAIK skipping deployment does fail the build in Brew. But I also noticed we were doing this in the other demos, maybe this will be OK. 

I'm not sure if there's some specifics to when Brew will and will not throw the unknown file in build root error.

AntStephenson: Hi @tiagobento, @psiroky,

I've taken a quick look at how we build this because it intrigued me. It turns out we're using PME to override the deploy plugin (http://release-engineering.github.io/pom-manipulation-ext/guide/plugin-manip.html#install-and-deploy-skip-flag-alignment). So in short, whatever you use here is fine. :)

tiagobento: @AntStephenson @psiroky Good! Thank you for looking into this :)

PR #274, opened 2017-07-30
Fix userCache proxy & avoid adding interfaces/abstracts as serializable types.

BenDol:
This catch is for wildcard serialization definitions (*).

If there is a better fix for these issues, I can update the PR.

Thanks

### DIFF
--- a/errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
+++ b/errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
@@ -302,8 +302,8 @@ public abstract class EnvUtil {
   }
 
   private static boolean validateWildcardSerializable(MetaClass mc) {
-    if(mc.isInterface() || (mc.isAbstract() && !mc.isEnum())) {
-      log.info("Serializable types cannot be an interface or abstract, ignoring: " + mc.getFullyQualifiedName());
+    if (mc.isInterface() || (mc.isAbstract() && !mc.isEnum())) {
+      log.debug("Serializable types cannot be an interface or abstract, ignoring: {}", mc.getFullyQualifiedName());
       return false;
     }
     return true;
--- a/errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
+++ b/errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
@@ -296,9 +296,17 @@ public abstract class EnvUtil {
       MetaClassFactory
         .getAllCachedClasses()
         .stream()
-        .filter(mc -> filter.apply(mc.getFullyQualifiedName()) && !mc.isInterface() && (!mc.isAbstract() || mc.isEnum()))
+        .filter(mc -> filter.apply(mc.getFullyQualifiedName()) && validateWildcardSerializable(mc))
         .collect(toCollection(() -> exposedClasses));
     }
+  }
+
+  private static boolean validateWildcardSerializable(MetaClass mc) {
+    if(mc.isInterface() || (mc.isAbstract() && !mc.isEnum())) {
+      log.info("Serializable types cannot be an interface or abstract, ignoring: " + mc.getFullyQualifiedName());
+      return false;
+    }
+    return true;
   }
 
   private static void updateFrameworkProperties(final Map<String, String> frameworkProps, final String key, final String value) {
--- a/errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
+++ b/errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
@@ -296,7 +296,7 @@ public abstract class EnvUtil {
       MetaClassFactory
         .getAllCachedClasses()
         .stream()
-        .filter(mc -> filter.apply(mc.getFullyQualifiedName()))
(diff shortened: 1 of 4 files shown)


tiagobento:
ok to test

errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
```diff
@@ -296,7 +296,7 @@ private static void addSerializableTypes(final Set<MetaClass> exposedClasses, fi
       MetaClassFactory
         .getAllCachedClasses()
         .stream()
-        .filter(mc -> filter.apply(mc.getFullyQualifiedName()))
+        .filter(mc -> filter.apply(mc.getFullyQualifiedName()) && !mc.isInterface() && (!mc.isAbstract() || mc.isEnum()))
```
mbarkley: Not sure I understand why we need to check if the type is an enum. If `mc.isEnum()` is true, then isn't it already the case that `!mc.isAbstract()` and `!mc.isInterface()`?

errai-security/errai-security-client/src/main/java/org/jboss/errai/security/client/local/context/SecurityContextImpl.java
```diff
@@ -111,6 +112,7 @@ public HTMLElement getElement() {
 
   @PostConstruct
   private void setup() {
+    userCache = Factory.maybeUnwrapProxy(userCache);
```
mbarkley: So why is it a problem here that the `userCache` is proxied? (If that is a case, a cleaner fix is to change the cache implementation from `@ApplicationScoped` to `@Singleton`).

errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
```diff
@@ -296,11 +296,19 @@ private static void addSerializableTypes(final Set<MetaClass> exposedClasses, fi
...
         .stream()
-        .filter(mc -> filter.apply(mc.getFullyQualifiedName()))
+        .filter(mc -> filter.apply(mc.getFullyQualifiedName()) && validateWildcardSerializable(mc))
         .collect(toCollection(() -> exposedClasses));
     }
   }
 
+  private static boolean validateWildcardSerializable(MetaClass mc) {
+    if(mc.isInterface() || (mc.isAbstract() && !mc.isEnum())) {
+      log.info("Serializable types cannot be an interface or abstract, ignoring: " + mc.getFullyQualifiedName());
```
mbarkley: We probably don't want to spam the logs with types that are non serializable. Could you please change this to debug level?

Also, to the unnecessary String concatenation when this log message isn't printed, you can do this:

    log.debug("Serializable types cannot be an interface or abstract, ignoring: {}", mc.getFullyQualifiedName());

errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
```diff
@@ -296,11 +296,19 @@ private static void addSerializableTypes(final Set<MetaClass> exposedClasses, fi
...
         .getAllCachedClasses()
         .stream()
-        .filter(mc -> filter.apply(mc.getFullyQualifiedName()))
+        .filter(mc -> filter.apply(mc.getFullyQualifiedName()) && validateWildcardSerializable(mc))
         .collect(toCollection(() -> exposedClasses));
     }
   }
 
+  private static boolean validateWildcardSerializable(MetaClass mc) {
+    if(mc.isInterface() || (mc.isAbstract() && !mc.isEnum())) {
```
mbarkley: Minor nit-pick, but could you please fix the formatting so there is a space between the `if` and first parenthesis?

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
```diff
@@ -467,4 +467,12 @@ public static boolean isBindableType(final MetaClass type) {
 
     return bindableTypes;
   }
+
+  private static boolean validateWildcard(MetaClass bindable) {
+    if(bindable.isFinal()) {
```
mbarkley: Please fix the formatting so there is space between the `if` and first parenthesis.

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
```diff
@@ -467,4 +467,12 @@ public static boolean isBindableType(final MetaClass type) {
...
+      log.info("@Bindable types cannot be final, ignoring: " + bindable.getFullyQualifiedName());
```
mbarkley: Same comment as the above logging: we probably don't want to spam the logs with non-bindable types. Could you please change this to debug level?

Also please see the above logging comment for how you can avoid the String concatenation when this message isn't printed.

BenDol:
I want to make it non-proxied because it was causing a problem for me authenticating. It was thinking that I was an anonymous user even though I had already logged in, so I would be redirected to the Unauthorized page every time I logged in. The cause was that the user cache was proxied and still contained the Anon user. When I inspected the proxy the real user was the target object.

PR #275, opened 2017-07-31
ERRAI-1068 - Extend EnabledByProperty annotation to allow multiple co…

cristianonicolai:
…nditions

### DIFF
--- a/errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/EnabledByProperty.java
+++ b/errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/EnabledByProperty.java
@@ -35,12 +35,28 @@ public @interface EnabledByProperty {
    * The name of the system property that enables or disables this bean.
    */
   String value();
-  
+
   /**
-   * Normally, the target bean is enabled if and only if the specified system
-   * property exists and has the value {@code "true"}. If {@code negated} is
-   * {@code true} then this sense is reversed: the bean is disabled if and only
-   * if the specified system property exists and has the value {@code "true"}.
+   * Normally, the target bean is enabled if the specified system property exists and
+   * matches the declared {@code matchValue}. If {@code negated} is {@code true}
+   * then this sense is reversed: the bean is disabled if the specified system property
+   * equals the {@code matchValue}.
    */
   boolean negated() default false;
+
+  /**
+   * When enabled, the matchValue is used as the default value in case the system property is not set.
+   */
+  boolean matchByDefault() default false;
+
+  /**
+   * Whether it should consider case sensitive when comparing the match value.
+   */
+  boolean caseSensitive() default false;
+
+  /**
+   * Expected property value to match for considering this bean enabled
+   */
+  String matchValue() default "true";
+
 }
--- a/errai-ioc/src/main/java/org/jboss/errai/ioc/rebind/ioc/bootstrapper/IOCProcessor.java
+++ b/errai-ioc/src/main/java/org/jboss/errai/ioc/rebind/ioc/bootstrapper/IOCProcessor.java
@@ -1315,21 +1315,24 @@ public class IOCProcessor {
 
   private boolean isEnabledByProperty(final MetaClass type) {
     final EnabledByProperty anno = type.getAnnotation(EnabledByProperty.class);
-    final boolean propValue = getPropertyValue(anno.value());
+    final boolean propValue = getPropertyValue(anno.value(),
+                                               anno.matchValue(),
+                                               anno.matchByDefault(),
+                                               anno.caseSensitive());
     final boolean negated = anno.negated();
 
     return propValue ^ negated;
   }
 
-  private boolean getPropertyValue(final String propName) {
+  protected boolean getPropertyValue(final String propName,
+                                     final String matchValue,
+                                     final boolean matchByDefault,
+                                     final boolean caseSensitive) {
     final String propertyValue = EnvUtil.getEnvironmentConfig().getFrameworkOrSystemProperty(propName);
     if (propertyValue == null) {
-      return false;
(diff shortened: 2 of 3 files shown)


errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/EnabledByProperty.java
```diff
@@ -43,4 +43,15 @@
    * if the specified system property exists and has the value {@code "true"}.
    */
   boolean negated() default false;
+
+  /**
+   * Default value in case the system property is not set at all.
+   */
+  String defaultValue() default "";
```
mbarkley: IMO this should be changed to a boolean because there are only two meaningful states:

1. Your `defaultValue` equals `matchValue`.
2. Your `defaultValue` does not equal `matchValue`.

I would suggest replacing this with

    boolean matchByDefault() default false;

where, when true, the default value of the system property will be the `matchValue` and `null` otherwise.

errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/EnabledByProperty.java
```diff
@@ -43,4 +43,15 @@
...
+
+  /**
+   * Expected property value to match for considering this bean enabled
+   */
+  String matchValue() default "true";
```
mbarkley: I wonder if we need to add a flag to ignore case as well? The issue is that we used to use `Boolean.parseBoolean`, which IIRC is case insensitive. If I'm remembering that correctly, it would be good to add something like

    boolean caseSensitive default false;

That way we would be completely backwards compatible, while still allowing case-sensitive properties to be used.

errai-ioc/src/test/java/org/jboss/errai/ioc/rebind/ioc/bootstrapper/IOCProcessorTest.java
```diff
@@ -0,0 +1,64 @@
+package org.jboss.errai.ioc.rebind.ioc.bootstrapper;
```
mbarkley: Missing the license header.

errai-ioc/src/test/java/org/jboss/errai/ioc/rebind/ioc/bootstrapper/IOCProcessorTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+
+@RunWith(MockitoJUnitRunner.class)
+public class IOCProcessorTest {
```
mbarkley: Since this currently only tests the property value, could you please rename it to something descriptive like `PropertyValueTest`?

cristianonicolai:
@mbarkley thanks for the review, hopefully, I've addressed it all, please take another look.

errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/EnabledByProperty.java
```diff
@@ -43,4 +43,20 @@
    * if the specified system property exists and has the value {@code "true"}.
    */
   boolean negated() default false;
+
+  /**
+   * When enabled, the current value is used as default value in case the system property is not set.
```
mbarkley: IMO, it's unclear what "current value" means here. I think it would be more clear as:

    When enabled, the matchValue is used as the default value in case the system property is not set.

errai-ioc/src/main/java/org/jboss/errai/ioc/client/api/EnabledByProperty.java
```diff
@@ -43,4 +43,20 @@
    * if the specified system property exists and has the value {@code "true"}.
```
mbarkley: We will need to update this javadoc. Now that we can match arbitrary properties, it isn't so much that `negated` matches `false` instead of `true`. A better description would be:

> Normally, the target bean is enabled if the specified system property exists and matches the declared {@code matchValue}. If {@code negated} is {@code true} then this sense is reversed: the bean is disabled if the specified system property equals the {@code matchValue}.

errai-ioc/src/main/java/org/jboss/errai/ioc/rebind/ioc/bootstrapper/IOCProcessor.java
```diff
@@ -1315,21 +1315,24 @@ private boolean isAlternativeEnabled(final MetaClass type) {
...
+                                     final boolean caseSensitive) {
     final String propertyValue = EnvUtil.getEnvironmentConfig().getFrameworkOrSystemProperty(propName);
-    if (propertyValue == null) {
-      return false;
-    }
-    try {
-      return Boolean.parseBoolean(propertyValue);
-    } catch (final Throwable t) {
-      throw new RuntimeException("Could not parse " + propName + " value, " + propertyValue + ", to boolean.", t);
+    if(propertyValue == null){
```
mbarkley: Minor nit-pick, but could you please fix the formatting on this line so that there is space around the parentheses?

cristianonicolai:
@mbarkley done :)

mbarkley:
@cristianonicolai Do you need this change for anything in the 7.x series (as opposed to just on master/8.x)? If not, I would prefer not to backport to 4.0.x today as @mbiarnes is attempting an Errai release.

cristianonicolai:
@mbarkley should be fine to leave it for 8.x for now. I can open a new PR in case I need it for 7.x.

PR #279, opened 2017-08-09
ERRAI-1070: MarshallerGeneratorFactory optimization for performance and js size.

slavap:


mbarkley:
ok to test

errai-codegen/src/main/java/org/jboss/errai/codegen/control/SwitchBlock.java
```diff
@@ -86,16 +87,20 @@ public String generate(Context context) {
...
+        if (!switchExprMetaClass.isAssignableFrom(value.getType().getErased())) {
           throw new InvalidTypeException(
               value.generate(context) + " is not a valid value for " + switchExprStmt.getType().getFullyQualifiedName());
         }
-        // case labels must be unqualified
         String val = value.generate(context);
-        final int idx = val.lastIndexOf('.');
-        if (idx != -1) {
-          val = val.substring(idx + 1);
+        if (!isStringSwitch) {
```
mbarkley: So if I am understanding this part correctly, this fixes possibly buggy behaviour when the case label is a string containing dots?

slavap: @mbarkley yes, fixes string case with dots.

PR #281, opened 2017-08-15
ERRAI-1065: Deprecate all classes and interfaces of Errai DOM wrappers

tiagobento:
https://issues.jboss.org/browse/ERRAI-1065


tiagobento:
@mbarkley Ok!

PR #288, opened 2017-09-06
ERRAI-1075: Add Errai APT Generators mechanism

tiagobento:
Description
---
Having the goal to replace all GWT generators by Annotation Processors, this PR creates a mechanism to generate code using the Java APT for `errai-bus` and `errai-jaxrs` initially.

This was achieved by using and improving the existing APT MetaClass implementation so that `errai-codegen` remains the library for generating code in Errai.

Two annotation processor projects were created. One of them to export relevant files (`errai-[module]-export-apt`) for an specific module and another to actually generate the code to be used by Web Apps (`errai-generators-apt`). An adicional module called `errai-common-apt` was created to share code between the two types of annotation processors.

`errai-bus` and `errai-jaxrs` both come with their own export files annotation processors as a transitive dependency, so that every module that depends on them is able to generate its own export files. Export Files are files which contain references to relevant types for an Errai module. [1]

>**NOTE:** The `errai-[module]-export-apt` modules require `maven-compiler-plugin` **3.2 or higher**.

Usage
---

Errai APT generators can be configured for a Web App in four steps:
1. Add an `@ErraiApp` annotation to any class on the project.
2. Create a profile named "apt-generators" that can be activated by the "apt-generators" property, includes the `errai-generators-apt` jar as a provided dependency and sets a property named "gwt.module.prefix" to "AptGenerators_". [2]
3. Create a module named `AptGenerators_[moduleName]` that inherits the actual module and overrides a property named "errai.useAptGenerators" with the value "true". (this property is defined in `ErraiCommonApt.gwt.xml`. `errai-bus` and `errai-jaxrs` inherit it and the other modules will too as they have its generators migrated as well. [3]
4. Add (or modify) the property `<module>` on the `gwt-maven-plugin` to be in the format `${gwt.module.prefix}ModuleName`.

The property `apt-generators` will then switch to APT generators instead of GWT generators. (e.g. `mvn clean install -Dapt-generators`)

>**NOTE:** Every module of your Web App should have been built generating its own Export Files.

Next steps
---
- Migrate the rest of Errai modules to use APT generators.
- Parse GWT module files to restrict visible classes to APT generators like the GeneratorContext does in GWT generators.
- Skip `@GenerateAsync` specific generators when APT generators are enabled.


errai-common-apt/src/main/java/org/jboss/errai/common/apt/generator/AbstractErraiModuleExportFileGenerator.java
```diff
@@ -0,0 +1,113 @@
...
+    final Set<APTClass> exportedTypes = exportFile.exportedTypes.stream()
+            .map(s -> new APTClass(s.asType()))
+            .collect(toSet());
+
+    final String className = ExportFileName.encodeAnnotationNameAsExportFileName(exportFile);
+    final String fullClassName = exportFilesPackagePath() + "." + className;
+    final ClassStructureBuilder<?> classBuilder = ClassBuilder.define(fullClassName).publicScope().body();
+
+    exportedTypes.forEach(
+            exportedType -> classBuilder.publicField(RandomStringUtils.randomAlphabetic(6), exportedType).finish());
```
mbarkley: So I think I understand why you are using random Strings here. Personally, my preference is to, wherever possible, use deterministic, stable, and human-readable identifiers.

This isn't a blocker for merging this PR, but I would like to improve on this in the future.

errai-common/src/main/java/org/jboss/errai/common/metadata/MetaDataScanner.java
```diff
@@ -142,7 +142,7 @@ private MetaDataScanner(final List<URL> urls, final File cacheFile) {
               final Class<?> aClass = Thread.currentThread().getContextClassLoader().loadClass(clsName);
               extensions.add(aClass.asSubclass(Vfs.UrlType.class));
             } catch (final Throwable t) {
-              log.warn("could not load class scanning extension: " + clsName, t);
+              throw new RuntimeException("could not load class scanning extension: " + clsName, t);
```
mbarkley: This change just caught my eye as it seems mostly unrelated. What was the motivation here?

errai-jaxrs/errai-jaxrs-client/src/main/java/org/jboss/errai/enterprise/rebind/MetaClassFinder.java
```diff
@@ -0,0 +1,35 @@
...
+import org.jboss.errai.codegen.meta.MetaClass;
+
+import java.lang.annotation.Annotation;
+import java.util.Collection;
+import java.util.function.BiFunction;
+
+/**
+ * @author Tiago Bento <tfernand@redhat.com>
+ */
+public interface MetaClassFinder
```
mbarkley: Should also probably be a `@FunctionalInterface`.

errai-bus/src/main/java/org/jboss/errai/bus/rebind/MetaClassFinder.java
```diff
@@ -14,24 +14,22 @@
...
 /**
- * An annotation for triggering a processor used for running unit tests.
- *
- * @author Max Barkley <mbarkley@redhat.com>
+ * @author Tiago Bento <tfernand@redhat.com>
  */
-@Retention(RUNTIME)
-@Target({ TYPE, FIELD, METHOD, PARAMETER })
-public @interface Hook {
+public interface MetaClassFinder
```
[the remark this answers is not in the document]
tiagobento: Oh yes, totally missed it.

errai-codegen/src/main/java/org/jboss/errai/codegen/util/AnnotationFilter.java
```diff
@@ -0,0 +1,32 @@
...
+import org.jboss.errai.codegen.meta.MetaAnnotation;
+
+import java.lang.annotation.Annotation;
+import java.util.Collection;
+import java.util.function.Function;
+
+/**
+ * @author Tiago Bento <tfernand@redhat.com>
+ */
+public interface AnnotationFilter extends Function<Collection<MetaAnnotation>, Collection<MetaAnnotation>> {
```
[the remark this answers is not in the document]
tiagobento: Same.

errai-common/src/main/java/org/jboss/errai/common/metadata/MetaDataScanner.java
tiagobento: This is a really good catch! I can't remember when or why I changed that. I'll revert it :)

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTClassUtil.java
```diff
@@ -0,0 +1,215 @@
...
+ */
+public final class APTClassUtil {
+
+  public static Types types;
+  public static Elements elements;
+
+  private APTClassUtil() {
+  }
+
+  public static void init(final Types types, final Elements elements) {
```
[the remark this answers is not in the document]
tiagobento: Yeah.. I mean, it works, but I'm not a huge fan of the design as well. Since this is the first PR of many, I'm positive we can address this issue in the future PRs :)

errai-jaxrs/errai-jaxrs-client/src/main/java/org/jboss/errai/enterprise/rebind/MetaClassFinder.java
tiagobento: Yes. I'll add it.

errai-common-apt/src/main/java/org/jboss/errai/common/apt/generator/AbstractErraiModuleExportFileGenerator.java
tiagobento: Agree 100%. That shouldn't be hard to change, I'll se what I can do.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTClass.java
```diff
@@ -14,17 +14,17 @@
  * limitations under the License.
  */
 
-package org.jboss.errai.codegen.apt;
+package org.jboss.errai.codegen.meta.impl.apt;
 
-import static org.jboss.errai.codegen.apt.APTClassUtil.elements;
```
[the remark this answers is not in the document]
tiagobento: Yeah.. absolutely. Thanks for mentioning that. Let's talk about it later, it shouldn't take long to solve this.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/AbstractHasAnnotations.java
```diff
@@ -16,19 +16,24 @@
...
+import static java.util.stream.Collectors.toList;
 
 /**
  * Contains shared functionality by all implementations of
  * {@link HasAnnotations}.
  * 
  * @author Christian Sadilek<csadilek@redhat.com>
  */
-public abstract class AbstractHasAnnotations implements HasAnnotations {
+public abstract class AbstractHasAnnotations implements HasAnnotations, HasMetaAnnotations {
```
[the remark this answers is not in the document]
tiagobento: At first I added methods to the `HasAnnotations` interface, but then I noticed there were places which looked not necessarily related to `errai-codegen` using the `HasAnnotations` interface. I wanted to make less impact possible on the existing code and get the first APT generators to work. If you have the time, could you take a closer look at the HasAnnotations interface?

tiagobento:
@mbarkley Thanks for your review! I replied your inline comments. The `APTMetaClassTest` was removed by accident. I'll make sure to bring it back :)

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/AbstractHasAnnotations.java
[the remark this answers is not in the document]
mbarkley: Is there any usage in particular that you saw that you were concerned about? I just did a search for `HasAnnotations` and it all looks code-generation related to me.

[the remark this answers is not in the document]
tiagobento: I was particularly concerned with the `HasNamedAnnotation` and the `DependencyGraphBuilder.Dependency` classes. I guess it was not obvious to me that they were indeed related to code-generation. Thanks for taking a look at that, I'll make sure to merge `HasMetaAnnotations` and `HasAnnotations`.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/HasAnnotations.java
```diff
@@ -53,4 +55,21 @@ default boolean unsafeIsAnnotationPresent(final Class<? extends Annotation> anno
             .findFirst()
             .orElse(null);
   }
+
+  default Optional<MetaAnnotation> getAnnotation(final Class<? extends Annotation> annotationClass) {
+    return Optional.ofNullable(unsafeGetAnnotation(annotationClass)).map(RuntimeMetaAnnotation::new);
+  }
+
+  @SuppressWarnings("unchecked")
```
mbarkley: Why did this change from `boolean` to `Boolean`?

tiagobento: I tend to use less primitives as possible.. I guess it's more like an old habit than anything. Since it's new code, I figured it was okay to change it. if you see any reason not to, I can change it back without problems.

mbarkley: It's fine, was just curious.

slavap: @tiagobento Just a remark - primitive types are better in client side code, because GWT compiler maps them to native js types if possible. Though it doesn't matter in this particular case.

tiagobento: @slavap Thanks for mentioning that. I think since GWT 2.8.0 Booleans are not boxed anymore. See under _JDK emulation_ section: http://www.gwtproject.org/release-notes.html#Release_Notes_2_8_0_BETA1. Anyway, primitives do perform better inside the JVM, so it might be the case I rethink my preferences :P

PR #290, opened 2017-09-15
ERRAI-1075: Create APT generator for errai-data-binding

tiagobento:
~Opening to trigger a full build in Jenkins.~

~I'm still working on an issue I discovered in `errai-data-binding` integrations tests code generation. An issue similar to [this one](https://issues.jboss.org/browse/ERRAI-512) happens on the APT MetaClass implementation for the TestModel class, so I'm investigating. This shouldn't impact result of tests though.~

**EDIT:** Fixed on 594175fde9aaf4aec7fe2a822d98b57782f3abfe

At the moment, Errai's integration tests are not using the code generated by APT generators. For the modules that have already been migrated, I enabled the `errai-generators-apt` module so that we have more use cases to catch issues like this one I described above.

The idea is to run every single integration test using both GWT and APT generators, since we'll support them coexisting for a while. This configuration can be done in a separate PR.

Description
---
- Migrate `errai-data-binding` module to generate `BindableProxyLoaderImpl` using APT generators too.
- Skip `@GenerateAsync` specific generators when APT generators are enabled.
- Create `@ErraiModule` annotation to replace the `ErraiApp.properties` configuration file in the APT generators environment.
- Initial restriction for exportable types to be under the package of an ErraiModule.

Next steps
---
- Parse GWT module files (`.gwt.xml`) to restrict visible classes to APT generators like the GeneratorContext does for GWT generators.
- Improve logging of APT generators (they're currently using `System.out`)
- Make all integration tests run using the code generated by both GWT and APT generators


errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTMethod.java
```diff
@@ -67,7 +67,7 @@ public MetaClass getReturnType() {
       return new APTClass(((Type.MethodType) typeMirror).getReturnType());
     }
 
-    return new APTClass(method.getReturnType());
+    return new APTClass(method.getReturnType()).getErased();
```
mbarkley: Do we know if all the MetaMethod implementations return an erased type?

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTParameter.java
```diff
@@ -49,7 +49,7 @@ public String getName() {
 
   @Override
   public MetaClass getType() {
-    return new APTClass(actualParameterType);
+    return new APTClass(actualParameterType).getErased();
```
mbarkley: Do we know if all the MetaParamter implementations return an erased type?

errai-common-apt/src/main/java/org/jboss/errai/common/apt/ErraiAptExportedTypes.java
```diff
@@ -73,55 +72,74 @@ public static void init(final Types types,
...
     return packageElement.getEnclosedElements()
             .stream()
-            .collect(groupingBy(ErraiAptExportedTypes::annotationName,
-                    flatMapping(ErraiAptExportedTypes::getExportedTypesFromExportFile, toSet())));
+            .collect(groupingBy(this::getAnnotationNameFromExportFileElement,
+                    flatMapping(this::getExportedTypesFromExportFileElement, toSet())));
+  }
+
+  private String getAnnotationNameFromExportFileElement(final Element e) {
+    return ExportFileName.decodeAnnotationClassNameFromExportFileName(e.asType().toString());
```
mbarkley: So a fun thing I learned last week is that the `toString()` on `TypeElement` is not standard between compiler implementations. In Eclipse it will give you a string like "class org.acme.Foo" instead of just the name. The portable way to do this is to call `e.asType().getQualifiedName().toString()`.

errai-data-binding/pom.xml
```diff
@@ -93,17 +93,33 @@
...
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-log4j12</artifactId>
       <scope>test</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.jboss.errai</groupId>
+      <artifactId>errai-data-binding-export-apt</artifactId>
+      <scope>compile</scope>
```
mbarkley: If you put this as compile scope, it will be transitively included in the classpath of webapps using data-binding. Is that intended or did you want to use provided?

tiagobento: It's intended. Every module, inside errai or not, that has errai-data-binding as a dependency should get errai-data-binding-export-apt as well, otherwise, its @Bindable types wouldn't be exported in an Export File and therefore not be visible to an Errai App.

errai-common-apt/src/main/java/org/jboss/errai/common/apt/ErraiAptExportedTypes.java
tiagobento: Oh, that's valuable info! Thanks, i'll totally change that :)

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTParameter.java
tiagobento: Yeah, it's a bit odd.. but GWTParameter and GWTMethod do return erased types. It may not seem ideal or even correct, but our code base expects that, so I think we should just copy the old behavior for now.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTMethod.java
tiagobento: Same as below.

errai-codegen/src/main/java/org/jboss/errai/codegen/MethodInvocation.java
```diff
@@ -95,8 +95,13 @@ else if (writer.getTypeParm(typeVar.getName()) != null) {
 
   // Resolves type variables by inspecting call parameters
   private void resolveTypeVariables() {
-    final MetaParameterizedType gSuperClass = inputType.getGenericSuperClass();
-    final MetaClass superClass = inputType.getSuperClass();
+    MetaParameterizedType gSuperClass = inputType.getGenericSuperClass();
+    MetaClass superClass = inputType.getSuperClass();
+
+    if (gSuperClass == null && superClass == null)  {
```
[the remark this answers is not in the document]
tiagobento: I'd have to debug everything again to remember why I added that. It certainly has something to do with generics, though. 

Just ran a quick build here without these lines and [this error](https://gist.github.com/tiagobento/92f3358dc2d8e3e3d1d4887b8659cc20) occurs while generating `BinableProxyLoaderImpl` for `errai-jpa-demo-grocery-list`.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTMethod.java
mbarkley: Okay, I'm assuming we don't have test coverage for this behaviour. Would you mind adding tests for this (and similar behaviour in MetaParameter) to the AbstractMetaClassTest?

tiagobento: There were tests for the return type of a MetaMethod, but not for MetaParameters. I added them with the same comment referencing [ERRAI-459](https://issues.jboss.org/browse/ERRAI-459). It seems like this behavior is there for a long time now.. This is certainly something that could be addressed in the future, but I don't see it as being a part of this migration we're in the middle of, so if it's okay for you, I'll not deal with that in this PR.

mbarkley: That's fine. I'm good to merge this PR when the build finishes.

mbarkley:
@tiagobento Looks like you need to add your new test class to the mock TypeOracle in the GWTMetaClassTest.

tiagobento:
@mbarkley Yeah.. just did that. Trying to figure out why didn't the GWTMetaClassTest run locally. I'll push it in a bit.

tiagobento:
@mbarkley Looks like you must activate the `integration-test` profile for it to run. Anyway, last commit should fix it :)

PR #292, opened 2017-10-05
AF-477: Support for Window's confirm method.

romartin:
Hey @mbarkley ,
As commented before, adding the `confirm` method in the Window object API because this needs to be back-ported to `7.3.x`/`4.0.x`. 
Thanks!


mbarkley:
@romartin I've cherry-picked and pushed to 4.0.x as well.

romartin:
thanks a lot @mbarkley !!

PR #293, opened 2017-10-11
ERRAI-1075 - Create APT generators for errai-ioc and errai-cdi

tiagobento:
This PR adds support for APT generators in the `errai-ioc` and `errai-cdi` modules.


errai-codegen/src/main/java/org/jboss/errai/codegen/meta/RuntimeAnnotation.java
```diff
@@ -63,4 +74,21 @@ public MetaClass annotationType() {
   public Annotation getAnnotation() {
     return annotation;
   }
+
+  @Override
+  public Map<String, Object> values() {
+    return Arrays.stream(annotation.annotationType().getDeclaredMethods())
+            .filter(m -> !m.isAnnotationPresent(Nonbinding.class))
```
mbarkley: Hmm, this code being here might be an issue. NonBinding is a CDI annotation, meaning that an attribute is not used to compare annotations for equality between observer, beans, and injection sites. In general, there could be places that care about NonBinding attributes.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/RuntimeAnnotation.java
```diff
@@ -16,25 +16,32 @@
...
 import java.util.Arrays;
+import java.util.Map;
+
+import static java.util.stream.Collectors.toMap;
 
 /**
  * @author Tiago Bento <tfernand@redhat.com>
  */
-public class RuntimeMetaAnnotation extends MetaAnnotation {
+public class RuntimeAnnotation extends MetaAnnotation {
```
mbarkley: Do you think we could change this to match the naming convention of the other MetaClass impls in this module and call this "JavaReflectionAnnotation"?

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/RuntimeEnum.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.errai.codegen.meta;
+
+/**
+ * @author Tiago Bento <tfernand@redhat.com>
+ */
+public class RuntimeEnum extends MetaEnum {
```
mbarkley: Same as with Annotation, I think a more consistent name would be "JavaReflectionEnum".

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTAnnotation.java
```diff
@@ -115,63 +110,31 @@ private static Object convertValue(final Object value, final Class<?> arrayType)
...
-    case VOID:
-      return void.class;
-    default:
-      return throwUnsupportedTypeError(value);
-    }
+  @Override
+  public Map<String, Object> values() {
+    return values.entrySet()
+            .stream()
+            .filter(e -> !isNonbinding(e.getKey()))
```
mbarkley: Same as before: the logic of filtering out NonBinding attributes probably should be external to this implementation.

errai-codegen/src/main/java/org/jboss/errai/codegen/util/GenUtil.java
```diff
@@ -480,10 +480,15 @@ else if (arguments.length == 0 && parmTypes.length == 0) {
 
           score = scoreMethods(arguments, parmTypes, isVarArgs);
 
+
           if (score != 0 && score > bestScore) {
             bestCandidate = meth;
             bestScore = score;
           }
+
+          if (score != 0 && score == bestScore && bestCandidate.isVarArgs()) {
```
mbarkley: Could you add a comment here explaining what corner case this is? Is this so we always prefer a var-args method over non-var-args when they are otherwise equally valid matches?

tiagobento: It's the contrary of that. We always prefer non-varargs methods over varargs ones. So at this point, when the bestCandidate is vararg, we switch. I'll add the comment.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/RuntimeAnnotation.java
tiagobento: Sure :)

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/RuntimeEnum.java
tiagobento: I'll rename this one too.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/RuntimeAnnotation.java
tiagobento: Hmm, I see. I'll place this logic of filtering out @Nonbinding methods outside the implementation of MetaAnnotations.

errai-codegen/src/main/java/org/jboss/errai/codegen/meta/impl/apt/APTAnnotation.java
tiagobento: Same as above.

errai-codegen/src/main/java/org/jboss/errai/codegen/util/GenUtil.java
mbarkley: Hehe yes I got confused there. All the more reason for the comment ;)

PR #294, opened 2017-10-19
ERRAI-1072: Add parameter errai.ui.nonbindableTypes to ErraiApp.properties

karandas210:


tiagobento:
ok to test

mbarkley:
Hi @karandas210, it looks like this PR breaks the nonPortableTypes property.  You can run this command in the errai-bus module to reproduce the failure:

    mvn clean install -Pintegration-test -Dit.test=SerializationTests

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
```diff
@@ -351,7 +351,8 @@ private static void assertTypeIsDataBinder(final MetaClass type) {
    *          the type to check
    */
   private static void assertTypeIsBindable(final MetaClass type) {
-    if (!type.isAnnotationPresent(Bindable.class) && !getConfiguredBindableTypes().contains(type)) {
+    if ((!type.isAnnotationPresent(Bindable.class) && !getConfiguredBindableTypes().contains(type))
```
mbarkley: Since this logic is getting more complex, it probably makes sense to reuse `isBindableType` here.

errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
```diff
@@ -210,7 +211,7 @@ private static void processErraiAppPropertiesBundle(final Map<String, String> fr
         addSerializableTypes(exposedClasses, explicitTypes, value);
       }
       else if (key.equals(CONFIG_ERRAI_NONSERIALIZABLE_TYPE)) {
-        addNonSerializableTypes(exposedClasses, nonportableClasses, value);
+          addNonSerializableTypes(nonportableClasses, value);
```
mbarkley: Please fix the indentation here.

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
```diff
@@ -413,43 +416,37 @@ public static boolean isBindableType(final MetaClass type) {
...
 
-  private static Set<MetaClass> findConfiguredBindableTypes() {
-    final Set<MetaClass> bindableTypes = new HashSet<>();
+  /**
+   * Reads nonbindable types from all ErraiApp.properties files on the classpath.
+   *
+   * @return a set of meta classes representing the configured nonbindable types.
+   */
+  public static Set<MetaClass> getConfiguredNonBindableTypes() {
+    if (configuredNonBindableTypes == null) {
```
mbarkley: In `getConfiguredBindableTypes` there is a "refresh" case when `configuredBindableTypes` has already been loaded. This is important for making things work properly during refreshes of Super Dev Mode. In this case you *might* be okay avoiding this step since we are explicitly not using non-bindable types to generate anything, but this should be tested.

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
```diff
@@ -464,8 +461,56 @@ public static boolean isBindableType(final MetaClass type) {
         }
       }
     }
+  }
 
-    return bindableTypes;
+  private static void addConfiguredBindableTypes(final Set<MetaClass> bindableClasses, String value) {
```
mbarkley: It seems the only difference between this method and the one below is the error message. Could we possibly refactor these to use a single helper method with an error message parameter?

karandas210:
Hi @mbarkley, I've already reproduced the failure. In my opinion, the cause of the problem is the test case. Could you take a look at the my changes in the addNonSerializableTypes method. Previously this method added classes founded by pattern in exposedClasses(using for generate marshallers) instead of excluding them. In my case the test failed because of another one exception("no marshalling definition available") raised in org.jboss.errai.marshalling.client.util.MarshallUtil.getMarshaller(Object, MarshallingSession). This worked fine earlier because the NonSerializable type does not contain any constructor or setters. So, Marshaller can't be generated in any case independently of the nonserializableTypes setting. I've checked it by adding default constructor to the NonSerializable type and Marshaller was generated and exception was not raised.

`public void testTypeInNonSerializablePackageIsNotPortable() throws Exception {
   try {
     Marshalling.toJSON(new NonSerializable());
   } catch (final RuntimeException ex) {
     assertTrue("Unexpected exception.", ex.getMessage().contains("No marshaller for type"));
   }
 }`

errai-config/src/main/java/org/jboss/errai/config/rebind/EnvUtil.java
karandas210: Ok

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
karandas210: Ok

errai-data-binding/src/main/java/org/jboss/errai/databinding/rebind/DataBindingUtil.java
```diff
@@ -464,8 +461,56 @@ public static boolean isBindableType(final MetaClass type) {
...
+          throw new RuntimeException("Could not find class defined in ErraiApp.properties as bindable type: " + s);
+        }
+      }
+    }
+    if (!patterns.isEmpty()) {
+      final SimplePackageFilter filter = new SimplePackageFilter(patterns);
+      MetaClassFactory
+        .getAllCachedClasses()
+        .stream()
+        .filter(mc -> filter.apply(mc.getFullyQualifiedName()) && validateWildcard(mc))
```
karandas210: It also difference in this place. validateWildcard(mc) check added to addConfiguredBindableTypes.

mbarkley: Okay. I personally would prefer still having a helper method with some kind of delegate to specify additional filtering, but this is not a blocker for me.

karandas210:
@mbarkley Ok, I will do that.

karandas210:
@mbarkley Finished.)) Please review.

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/nonbindablepkg/TestModelInNonBindablePkg.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (C) 2011 Red Hat, Inc. and/or its affiliates.
```
mbarkley: Please update the year.

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/test/DataBindingIntegrationTest.java
```diff
@@ -1588,4 +1589,14 @@ public void testDeclarativeBindingWithOnKeyUpForTextInput() throws Exception {
     invokeEventListeners(view.value, "keyup");
     assertEquals(view.value.getValue(), model.getValue());
   }
+
+  @Test
+  public void testTypeInNonBiandablePakageIsNotBound() {
```
mbarkley: Minor thing, but there are a couple typos here. Should probably be `testTypeInNonBindablePackageIsNotBound`.

errai-data-binding/src/test/java/org/jboss/errai/databinding/client/test/DataBindingIntegrationTest.java
```diff
@@ -1588,4 +1589,14 @@ public void testDeclarativeBindingWithOnKeyUpForTextInput() throws Exception {
...
+    try {
```
mbarkley: Thank you for adding this test! One thing I've noticed is that I think this test would succeed if type were bindable (i.e. if not exception is thrown).

Here is a pattern I've used in other tests that I think would be useful here:

    try {
      somethingThatShouldHaveAnError();
      fail("Something succeeded but should have failed.");
    } catch (AssertionError ae) {
      throw ae;
    } catch (Exception ex) {
      assertTrue(someCondition(ex));
    }

PR #298, opened 2017-11-07
Add errai.bus.web_socket_max_frame_size property

BenDol:
Required to expand the default websocket frame size.

Pretty simple change, please let me know if there is a better implementation.

Thanks!


mbarkley:
Merged. Thanks, @BenDol!

PR #300, opened 2017-11-09
Upgrade ip bom

mbarkley:
Fixes version incompatibilities that lead to test failures in errai-bus-jsr356-websocket-weld.

### DIFF
--- a/errai-bom/pom.xml
+++ b/errai-bom/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <groupId>org.jboss.integration-platform</groupId>
     <artifactId>jboss-integration-platform-parent</artifactId>
-    <version>7.0.0.CR9</version>
+    <version>8.0.0.CR2</version>
     <relativePath/>
   </parent>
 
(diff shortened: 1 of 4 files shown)


errai-bom/pom.xml
```diff
@@ -47,7 +47,7 @@
     <javaee.api.version>7.0</javaee.api.version>
     <jboss-integration-platform-bom.version>${project.parent.version}</jboss-integration-platform-bom.version>
     <!-- ip-bom uses artifact with groupId org.jboss.remoting3 -->
-    <jboss.remoting.version>4.0.0.Final</jboss.remoting.version>
+    <jboss.remoting.version>5.0.3.Final</jboss.remoting.version>
```
mbarkley: @psiroky @mbiarnes This version ended up causing most of my issues. The ip-bom manages a version for jboss-remoting 3, which uses a separate package and groupId (org.jboss.remoting3). But wildfly-core depends on a newer jboss-remoting.

Should we be managing the version for org.jboss.remoting as well in the ip-bom?

psiroky: WildFly seems to be using `org.jboss.remoting:jboss-remoting:5.0.3.Final`, so I guess we should get rid of jboss-remoting3 and move the 5.0.3.Final version to ip-bom.

mbarkley: @psiroky Okay. For now I will merge this because we depend on wildfly-cli directly in errai-cdi-jboss so I need to have a managed version. We can get rid of this when the ip-bom is upgraded.

PR #303, opened 2017-11-14
Add Elemental2DomUtil (utility methods for interacting with the DOM)

karreiro:
Regarding the `Elemental2DomUtil`, it's an Elemental2 approach for the deprecated `DOMUtil` (https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/dom/DOMUtil.java).

Regarding the `errai-internal-bom/pom.xml`, the `elemental2-dom` artifact is being added to avoid classes referenced via transitive dependencies. Since we're deprecating Errai elements in favor of Elementa2, we should make them available `errai-internal-bom/pom.xml`.

### DIFF
--- a/errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
+++ b/errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.errai.common.client.dom.elemental2;
+
+import com.google.gwt.user.client.ui.RootPanel;
+import com.google.gwt.user.client.ui.Widget;
+import elemental2.dom.HTMLElement;
+import elemental2.dom.Node;
+import jsinterop.base.Js;
+
+/**
+ * Provides utility methods for interacting with the DOM.
+ * @author Guilherme Carreiro <ggomes@redhat.com>
+ */
+public class Elemental2DomUtil {
+
+  /**
+   * Detaches all element children from a node.
+   * @param node Must not be null.
+   * @return True iff any element children were detached by this call.
+   */
+  public boolean removeAllElementChildren(final Node node) {
+
+    final boolean hadChildren = node.lastChild != null;
+
+    while (node.lastChild != null) {
+      node.removeChild(node.lastChild);
+    }
+
+    return hadChildren;
+  }
+
+  /**
+   * Appends the underlying {@link HTMLElement} of a {@link Widget} to another {@link HTMLElement},
+   * in a way that does not break GWT Widget events.
+   * @param parent The parent element that is appended to. Must not be null.
+   * @param child The child Widget, whose underlying HTML element will be
+   * appended to the parent. Must not be null.
+   */
+  public void appendWidgetToElement(final HTMLElement parent, final Widget child) {
+
+    if (child.isAttached()) {
+      child.removeFromParent();
+    }
+
+    RootPanel.detachOnWindowClose(child);
+
+    HTMLElement newChild = asHTMLElement(child.getElement());
+    parent.appendChild(newChild);
+
+    onAttach(child);
+  }
+
+  /**
+   * Converts from {@link com.google.gwt.dom.client.Element} to {@link HTMLElement}.
+   * @param gwtElement The deprecated GWT Element
+   * @return The GWT Elemental2 HTMLElement
+   */
+  public HTMLElement asHTMLElement(final com.google.gwt.dom.client.Element gwtElement) {
+    return Js.cast(gwtElement);
+  }
+
+  /**
+   * Converts from {@link org.jboss.errai.common.client.dom.HTMLElement} to {@link HTMLElement}.
+   * @param htmlElement The deprecated Errai HTMLElement
+   * @return The GWT Elemental2 HTMLElement
+   */
+  public HTMLElement asHTMLElement(final org.jboss.errai.common.client.dom.HTMLElement htmlElement) {
+    return Js.cast(htmlElement);
+  }
+
+  native void onAttach(Widget w)/*-{
+    w.@com.google.gwt.user.client.ui.Widget::onAttach()();
+  }-*/;
+}
--- a/errai-common/src/test/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtilTest.java
+++ b/errai-common/src/test/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtilTest.java
@@ -1,0 +1,211 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jboss.errai.common.client.dom.elemental2;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import com.google.gwt.junit.GWTMockUtilities;
+import com.google.gwt.user.client.Element;
+import com.google.gwt.user.client.ui.RootPanel;
+import com.google.gwt.user.client.ui.Widget;
+import elemental2.dom.HTMLElement;
+import elemental2.dom.Node;
+import jsinterop.base.Js;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+import static org.powermock.api.mockito.PowerMockito.verifyStatic;
+
+/**
+ * Testing Elemental2DomUtil API.
+ * @author Guilherme Carreiro <ggomes@redhat.com>
+ */
+@RunWith(PowerMockRunner.class)
+@PrepareForTest({ Elemental2DomUtil.class, RootPanel.class, Js.class })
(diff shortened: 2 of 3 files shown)


karreiro:
@manstis @mbarkley Could you please take a look? :-)

errai-internal-bom/pom.xml
```diff
@@ -302,6 +306,12 @@
         <artifactId>errai-api</artifactId>
         <version>${project.version}</version>
       </dependency>
+
+      <dependency>
+        <groupId>com.google.elemental2</groupId>
+        <artifactId>elemental2-dom</artifactId>
+        <version>${version.com.google.elemental2}</version>
+      </dependency>
```
tiagobento: I think the place to add this dependency would be in `errai-bom` instead of `errai-internal-bom`. `errai-internal-bom` is supposed to contain the errai modules only. IIRC the elemental2 dependency is already listed on `errai-bom`.

karreiro: Thanks for the comment @tiagobento :-) However, I think that the `errai-bom` and the `errai-internal-bom` have different responsibilities.

Many projects using Errai import the `errai-internal-bom` ([drools-wb](https://github.com/kiegroup/drools-wb/blob/master/pom.xml#L73), [jbpm-wb](https://github.com/kiegroup/jbpm-wb/blob/master/pom.xml#L75), [dashbuilder](https://github.com/dashbuilder/dashbuilder/blob/master/pom.xml#L545), [kie-wb-distribution](https://github.com/kiegroup/kie-wb-distributions/blob/master/pom.xml#L66)).\

That allows the users of the Errai API to declare Errai modules and avoid illegal transitive dependencies ([see](https://github.com/kiegroup/drools-wb/blob/master/drools-wb-screens/drools-wb-guided-dtable-editor/drools-wb-guided-dtable-editor-client/pom.xml#L72)).

But now, that the Errai elements are deprecated, sounds fair enough to get the Elemental2 **form** Errai. I mean, I don't think that we should add the Elemental2 version in the `drools-wb` (or even, in `droolsjbpm-build-bootstrap`).

WDYT?

karreiro:
Jenkins please retest this.

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
```diff
@@ -0,0 +1,99 @@
...
+
+import com.google.gwt.user.client.ui.RootPanel;
+import com.google.gwt.user.client.ui.Widget;
+import elemental2.dom.HTMLElement;
+import elemental2.dom.Node;
+import jsinterop.base.Js;
+
+/**
+ * Provides utility methods for interacting with the DOM.
+ * @author Guilherme Carreiro <ggomes@redhat.com>
```
manstis: Our preferred _style_ is to not include ```@author``` tags.

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
```diff
@@ -0,0 +1,99 @@
...
+        final Elemental2Widget widget = Js.uncheckedCast(child);
+
+        widget.onAttach();
+    }
+
+    static class Elemental2Widget extends Widget {
+
+        @Override
+        protected void onAttach() {
+            super.onAttach();
```
manstis: I'm not understanding the significance of overriding this method!? 

If it has some significance perhaps a comment in the code would be useful?

errai-common/src/test/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtilTest.java
```diff
@@ -0,0 +1,214 @@
...
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+import static org.powermock.api.mockito.PowerMockito.verifyStatic;
+
+/**
+ * Testing Elemental2DomUtil API.
+ * @author Guilherme Carreiro <ggomes@redhat.com>
```
manstis: Our preferred _style_ is to not include ```@author``` tags.

errai-internal-bom/pom.xml
tiagobento: Yeah.. I didn't know about this transitive dependencies check you guys have. I guess in this case it makes sense to have this dependency on the `errai-internal-bom`, but then I'd remove it from the `errai-bom`, given that then it would be duplicated inside Errai. 

I'd wait to hear what other people have to say about this matter, though.

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
karreiro: This is a workaround to access the `onAttach` protected method without JSNI.

[Here](https://github.com/errai/errai/blob/master/errai-common/src/main/java/org/jboss/errai/common/client/dom/DOMUtil.java#L439-L441) the previous version :-)

errai-common/src/test/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtilTest.java
karreiro: I think that Errai adopts a different code style.. Am I missing something?

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
karreiro: I think that Errai adopts a different code style.. Am I missing something?

errai-common/src/test/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtilTest.java
manstis: Probably I am!

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
manstis: Probably I am!

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
```diff
@@ -0,0 +1,99 @@
...
+     * Converts from {@link org.jboss.errai.common.client.dom.HTMLElement} to {@link HTMLElement}.
+     * @param htmlElement The deprecated Errai HTMLElement
+     * @return The GWT Elemental2 HTMLElement
+     */
+    public static HTMLElement asHTMLElement(final org.jboss.errai.common.client.dom.HTMLElement htmlElement) {
+        return Js.cast(htmlElement);
+    }
+
+    private static void attach(final Widget child) {
+        final Elemental2Widget widget = Js.uncheckedCast(child);
```
mbarkley: To be honest, this makes me uneasy. As far as I know, `Js.cast` and `Js.uncheckedCast` are meant for casting between native JavaScript types, not GWT/Java types. How have you tested this? Does it work in a production compiled showcase?

My concern here is that in certain compilations (like a production compile with full optimizations) this cast might succeed but the method call to onAttach might fail because the underlying object is not the correct type.

mbarkley:
Jenkins, please retest this.

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
```diff
@@ -0,0 +1,99 @@
...
+public class Elemental2DomUtil {
+
+    /**
+     * Detaches all element children from a node.
+     * @param node Must not be null.
+     * @return True iff any element children were detached by this call.
+     */
+    public static boolean removeAllElementChildren(final Node node) {
+
+        final boolean hadChildren = node.lastChild != null;
```
jomarko: Maybe could be more defensive: `final boolean hadChildren = node != null && node.lastChild != null;`, not sure let me know what do you think.

mbarkley: I don't think it's a good idea to silently return if a null is passed in. We could enhance it to assert the parameter is non-null, but it's not a big deal for me since a null would cause an NPE on this line where it would be obvious that node was the culprit.

jomarko: @mbarkley thanks for you comment, it sounds reasonable to me, just to explain my point a little bit more, I wanted to prevent NPE because it is public util method, which doesn't mention this NPE behavior in javadoc, that is why I thought NPE shouldn't be thrown so easy, @karreiro leaving the decision on you, if you prefer variant with NPE, could we have mentioned this in javadoc?

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
karreiro: @mbarkley Yeah, these casts are a little bit nasty. But, unfortunately, I see them as the solution with least impact that we have to solve this problem right now.

I tested that on a production compiled showcase, and the components work as expected. The following Gif shows a component (Accordion) that depends of the `Elemental2DomUtil`:

[gif - interagindo com accordion] 1

Since the signature of the `attach` method is:
```java
void attach(final Widget child)`
````
...and the `Widget` has a protected method called `onAttach`, I don't think that we're going to see a `child` without the `onAttach` method being called.

Am I missing something?

Thanks.

karreiro:
Thank you for the valuable review @mbarkley. Regarding the topics that you raised:
- 2-spaces for indentation: fixed;
- errai-internal-bom: fixed;
- injectable-dom-util: fixed;
- concerns about `Widget.onAttach`: answered (in-line).

I hope to have covered everything that you pointed out.

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
karreiro: Thanks for the comment @jomarko.

However, I agree with @mbarkley, a NullPointerException on this line is clear enough.

I prefer to keep it as is, to keep it simple (or simpler).

mbarkley: FWIW the javadoc looks good to me, because it already states that the node parameter must not be null. Saying that a parameter must not be null makes it fairly obvious what the problem was when you get an NPE.

errai-common/src/main/java/org/jboss/errai/common/client/dom/elemental2/Elemental2DomUtil.java
mbarkley: So I gave this a try in the Errai tutorial, and there is an error calling attach when production compiled. I suspect this screenshot was from one of the webapps that is configured to use draft compile, even when you do a `mvn clean package` (without the "full" profile).

[Here](https://github.com/mbarkley/errai-tutorial/tree/unchecked-cast-repro) is a reproducer in the errai-tutorial. Just run `mvn clean package`, deploy the app to WildFly, open the app, and look in the console for the log message "Something went terribly wrong".

The fix, I think, is to change the attach method back to JSNI. This is an instance where JSNI is the only solution until we are ready to migrate to GWT3 and have a GWT3-specific replacement.

karreiro: @mbarkley I recorded the Gif using a war from `kie-wb-distributions` (master) on WildFly 10.1.0.

But, I'm going to change it to JSNI for now :-) Thanks for the tip.

karreiro:
@mbarkley Changes applied.

I tested with the [reproducer](https://github.com/mbarkley/errai-tutorial/tree/unchecked-cast-repro) that you provided, and with `kie-wb-distributions` (master with `-Dfull`). Everything is working as expected.

PR #304, opened 2017-11-14
Fix issues where deferred messages error handling would be ignored.

BenDol:
This accounts for times where messages are ignored, rather than sink them, we should be throwing a NoSubscribersToDeliverTo exception.

Let me know if there is a more acceptable way to handle this.

### DIFF
--- a/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
+++ b/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
@@ -948,25 +948,44 @@ public class ClientMessageBusImpl implements ClientMessageBus {
     final List<Message> highPriority = new ArrayList<>();
     for (final Message message : new ArrayList<>(deferredMessages)) {
       if (message.hasPart(MessageParts.PriorityProcessing)) {
-        if (remotes.containsKey(message.getSubject()))
+        if (remotes.containsKey(message.getSubject())) {
           highPriority.add(message);
-        deferredMessages.remove(message);
+          deferredMessages.remove(message);
+        }
       }
     }
 
     final List<Message> lowPriority = new ArrayList<>();
-    do {
+    for(Message m : deferredMessages) {
       for (final Message message : new ArrayList<>(deferredMessages)) {
-        if (remotes.containsKey(message.getSubject()))
+        if (remotes.containsKey(message.getSubject())) {
           lowPriority.add(message);
-        deferredMessages.remove(message);
+          deferredMessages.remove(message);
+        }
       }
     }
-    while (!deferredMessages.isEmpty());
 
-    transportHandler.transmit(highPriority);
-    transportHandler.transmit(lowPriority);
-    deferredMessages.clear();
+    try {
+      transportHandler.transmit(highPriority);
+      transportHandler.transmit(lowPriority);
+
+      for (final Message message : deferredMessages) {
+        String subject = message.getSubject();
+        if (!localSubscriptions.containsKey(subject) ||
+            !subscriptions.containsKey(subject) ||
+            !shadowSubscriptions.containsKey(subject)) {
+          try {
+            throw new NoSubscribersToDeliverTo(subject);
+          } catch (NoSubscribersToDeliverTo ex) {
+            if (message.getErrorCallback() != null) {
+              message.getErrorCallback().error(message, ex);
+            }
+          }
+        }
+      }
+    } finally {
+      deferredMessages.clear();
+    }
   }
 
   public boolean handleTransportError(final BusTransportError transportError) {
--- a/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
+++ b/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
@@ -956,12 +956,10 @@ public class ClientMessageBusImpl implements ClientMessageBus {
     }
 
     final List<Message> lowPriority = new ArrayList<>();
-    for(Message m : deferredMessages) {
-      for (final Message message : new ArrayList<>(deferredMessages)) {
-        if (remotes.containsKey(message.getSubject())) {
-          lowPriority.add(message);
-          deferredMessages.remove(message);
-        }
+    for (final Message message : new ArrayList<>(deferredMessages)) {
+      if (remotes.containsKey(message.getSubject())) {
+        lowPriority.add(message);
+        deferredMessages.remove(message);
       }
     }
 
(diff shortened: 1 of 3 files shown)


tiagobento:
ok to test

mbarkley:
Rerunning the tests because a couple of them were from problems on master that should be fixed now.

Jenkins, please retest this.

BenDol:
Ah okay, will take a look at that.

mbarkley:
@BenDol I took a look before sending my last message and it appears that messages are sent to local subscribers immediately, even if the bus is not connected and the message is also deferred for remote subscribers. Personally, I don't like that, but changing that is out of scope for this fix IMO.

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
```diff
@@ -945,28 +945,53 @@ private void sendAllDeferred() {
...
-    transportHandler.transmit(highPriority);
-    transportHandler.transmit(lowPriority);
-    deferredMessages.clear();
+    try {
+      transportHandler.transmit(highPriority);
+      transportHandler.transmit(lowPriority);
+
+      for (final Message message : localMessages) {
+        String subject = message.getSubject();
+        if(!localSubscriptions.containsKey(subject)) {
```
mbarkley: This is almost there, but I think we need to change this condition to this:

    !(localSubscriptions.containsKey(subject) 
      || subscriptions.containsKey(subject)
      || shadowSubscriptions.containsKey(subject))

If you look in the implementation of the send method in the bus. Those three fields + remotes are all the places where a message can be delivered, so if any of those three contains a matching subject then the message had subscribers.

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
```diff
@@ -945,28 +945,53 @@ private void sendAllDeferred() {
     if (!deferredMessages.isEmpty())
       logger.info("transmitting deferred messages now ...");
 
+    List<Message> localMessages = new ArrayList<>();
```
mbarkley: So be low it looks like we are already removing delivered message from `deferredMessages` as we process them. Could we possibly clean this up a bit by getting rid of this field and iterating through the remaining `deferredMessages` instead?

BenDol: Good idea, will just have to replace the `while` loop with a `for` loop.

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
BenDol: Ah okay, wasn't sure if they were all required. Thanks! Will make the changes now.

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
```diff
@@ -948,25 +948,44 @@ private void sendAllDeferred() {
...
           highPriority.add(message);
-        deferredMessages.remove(message);
+          deferredMessages.remove(message);
+        }
       }
     }
 
     final List<Message> lowPriority = new ArrayList<>();
-    do {
+    for(Message m : deferredMessages) {
```
mbarkley: I don't think we need this outer loop anymore, and in fact, it is causing a ConcurrentModificationException (because of the deferredMessage.remove call) that is the source of some test failures.

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
```diff
@@ -948,25 +948,44 @@ private void sendAllDeferred() {
...
-    transportHandler.transmit(highPriority);
-    transportHandler.transmit(lowPriority);
-    deferredMessages.clear();
+    try {
+      transportHandler.transmit(highPriority);
+      transportHandler.transmit(lowPriority);
+
+      for (final Message message : deferredMessages) {
+        String subject = message.getSubject();
+        if (!localSubscriptions.containsKey(subject) ||
```
mbarkley: The condition here is not quite right. What I had intended was:

    !(localSubscriptions.containsKey(subject)
      || subscriptions.containsKey(subject)
      || shadowSubscriptions(subject))

Or in English: There is no subscription in any of local subscriptions, subscriptions, or shadow subscriptions.

I think fixing this condition will fix the remainder of the test failures.

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
BenDol: True you are right, lack of sleep makes for A+ coding haha, will fix that up asap

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
BenDol: 👍

tiagobento:
@BenDol So I took a look at the failing tests and they need to be changed a bit.

These tests are breaking because they test the exact same scenario this PR fixes. They send a deferred message to a non-responding service and expect a no-op to happen. But now, we have to make sure that, in such cases, the error callback is called with the right exception and message.

`org.jboss.errai.bus.client.tests.ServiceAnnotationTests.testMethodWithLocalService1`
`org.jboss.errai.bus.client.tests.ServiceAnnotationTests.testClassWithLocalService1`
`org.jboss.errai.cdi.event.client.test.CDIServiceAnnotationTests.testMethodWithLocalService1`
`org.jboss.errai.cdi.event.client.test.CDIServiceAnnotationTests.testClassWithLocalService1`

After this change, I think this PR is good to go :)

tiagobento:
@BenDol Thanks for taking the time for fixing the tests. It looks almost ready.

Before merging, though, I'd like to see assertions for the exception and message inside the error callback. Also, we need to make sure that, if a timeout happens, the tests fail. There are essentially three cases possible after sending a deferred message in these test cases:

1. Nothing happened. (That's when the timeout happens. Tests should fail in this case)
2. Message is sent. (Tests should fail in this case)
3. An error happened. (Here, we should include assertions for which error happened. In this case we are expecting a `NoSubscribersToDeliverTo` exception to happen).

BenDol:
In a `runNonRespondingServiceTest` it is expected to be failing though, right?

BenDol:
I keep getting: `junit.framework.AssertionFailedError: No response received after 10006 ms` when I run the tests locally no idea why. I am not able to run the tests.

tiagobento:
@BenDol The timeout was expected before this PR. Now, if a timeout occurs, we know that the error didn't happen, and so the test must fail.

Which SO are you running the tests on and how are you running them? If you're not able to run them locally at the moment, you can commit your changes and let Jenkins do the job :)

tiagobento:
@BenDol Changes look good to me, the only thing that's making the tests fail right now is that you used `Assert.assertTrue` and it can't be used inside a `GWTTestCase`. Changing it to `assertTrue` from the `TestCase` class should fix the problem.

mbarkley:
Thanks @BenDol!

PR #306, opened 2017-11-15
[AF-549-elastaic]: upgraded jackson version to 2.8.8

adrielparedes:
It seems that I also modified some spaces into pom.xml, If those shouldn't be there tell me and I will revert them.

@mbarkley Can you review this please?


porcelli:
Jenkins execute full downstream build

PR #310, opened 2017-11-20
Fixed issues where DefaultErrorHandlers wouldn't be invoked in RpcProxy's

BenDol:
The previous generation:
```java
        } catch (Throwable throwable) {
          if (errorCallback != null) {
            errorCallback.error(null, throwable);
          } else {
            invokeDefaultErrorHandlers(throwable);
          }
        }
```

Becomes:
```java
         } catch (Throwable throwable) {
          if (errorCallback != null) {
            if (errorCallback.error(null, throwable)) {
              invokeDefaultErrorHandlers(throwable);
            }
          } else {
            invokeDefaultErrorHandlers(throwable);
          }
        }
```

No longer ignoring the return type or `errorCallback.error`.

### DIFF
--- a/errai-bus/src/main/java/org/jboss/errai/bus/rebind/RpcProxyGenerator.java
+++ b/errai-bus/src/main/java/org/jboss/errai/bus/rebind/RpcProxyGenerator.java
@@ -16,14 +16,11 @@
 
 package org.jboss.errai.bus.rebind;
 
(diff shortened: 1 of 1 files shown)


mbarkley:
ok to test

mbarkley:
Thanks @BenDol

PR #311, opened 2017-11-20
Support custom security exception handling

BenDol:
This fixes a security issue with ClientSecurityRoleInterceptor if an invalid user cache is found it simply proceeds the call, however we should check if there are required roles, if so, then we should not proceed, but instead throw an UnauthenticatedException. This lead to simply making the call before the user was validated and it would succeed.

Use a SecurityExceptionHandler to define SecurityException handling or stick with the default.

Example of using your own SecurityExceptionHandler:
```java
@Singleton
@Alternative
public class SecurityExceptionHandler extends DefaultSecurityExceptionHandler {

    @Inject
    public SecurityExceptionHandler(SecurityContext context) {
        super(context);
    }

    @Override
    public boolean handleException(Throwable caught) {
        // We will only allow for redirection if we are in the
        // navigation phase.
        return !context.getNavigation().isNavigating() || super.handleException(caught);
    }
}
```

ErraiApp.properties:
```
errai.ioc.enabled.alternatives=xx.yy.zz.MySecurityExceptionHandler
```

### DIFF
--- a/errai-cdi/errai-cdi-client/src/main/java/org/jboss/errai/enterprise/client/cdi/InstanceProvider.java
+++ b/errai-cdi/errai-cdi-client/src/main/java/org/jboss/errai/enterprise/client/cdi/InstanceProvider.java
@@ -89,7 +89,7 @@ public class InstanceProvider implements ContextualTypeProvider<Instance> {
     @Override
     public Iterator<T> iterator() {
       final Collection<SyncBeanDef<T>> beanDefs = IOC.getBeanManager().lookupBeans( type, qualifiers );
-      if(beanDefs==null){
+      if (beanDefs==null){
         return Collections.<T>emptyList().iterator();
       }
       return new InstanceImplIterator(beanDefs);
--- a/errai-cdi/jboss/src/main/java/org/jboss/errai/cdi/server/gwt/EmbeddedWildFlyLauncher.java
+++ b/errai-cdi/jboss/src/main/java/org/jboss/errai/cdi/server/gwt/EmbeddedWildFlyLauncher.java
@@ -158,11 +158,11 @@ public class EmbeddedWildFlyLauncher extends ServletContainerLauncher {
 
         for (final String currentLine : lines) {
           String trimmed = currentLine.trim();
-          if(trimmed.startsWith(ERRAI_PROPERTIES_HINT_START.trim())) {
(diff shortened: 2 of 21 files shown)


mbarkley:
ok to test

BenDol:
Well I agree, but isn't the point of `@RestrictedAccess` to state that the call should not be executed without the given access loaded on the client and then server restrictions also be applied. Does the `@RestrictedAccess` support server side verification too? I am using Spring without the Weld CDI booting up so maybe that is why it wasn't properly handled?

mbarkley:
The point of `@RestrictedAccess` is to declare role-based security constraints. When these constraints are on client-side resources (@Pages) they are enforced on the client. When they are on server-side resources (RPC endpoints), they are meant to be enforced on the server. The interceptor is only an optimisation, to avoid unnecessary round trips when we know a call will fail.

The branch you changed in the interceptor is hit if the security context cache is invalid. That is often, but not always, the same as being logged out.

BenDol:
If we use the `@Alternative` way, do the users have to add that to the bean manager manually? Or does the container detect the alternative and replace the default automatically?

mbarkley:
Alternatives have to enabled in ErraiApp.properties.

BenDol:
Is there any documentation that I can look at for this?

mbarkley:
So the [DefaultErrorCallback](https://github.com/errai/errai/blob/master/errai-security/errai-security-client/src/test/java/org/jboss/errai/mocksafe/test/DefaultErrorCallbackTest.java) mockito tests are broken.

Basically there are 5 things being tested with two implementations under test:

1. UnauthenticatedException redirects to login page
2. UnauthorizedException redirects to security error page
3. Other errors are ignored
4. Redirecting to login throws MissingPageException when no login page is present
5. Redirecting to security error page throws MissingPageException when no security error page is present

The best solution is probably to replace these 2x5 tests with 5 tests directly against the default security handler implementation, plus two new tests that assert that the bus and rest error callbacks hall the security handler with the given exception.

BenDol:
Why are the tests failing? Is it because the `DefaultSecurityExceptionHandler` isn't mocked, since it is the class calling  `redirectToLoginPage` and `redirectToSecurityErrorPage`?

mbarkley:
Thanks @BenDol!

PR #313, opened 2017-11-21
Upgrade GWT to 2.8.2

psiroky:
GWT 2.8.2 is the first release that should support bulding with Java 9
(it does not support Java 9 features; just the legacy classpath mode)

### DIFF
--- a/errai-bom/pom.xml
+++ b/errai-bom/pom.xml
@@ -41,7 +41,7 @@
     <guava.version>20.0</guava.version>
     <guice.version>3.0</guice.version>
     <gwt.phonegap.version>3.5.0.1</gwt.phonegap.version>
-    <gwt.version>2.8.1</gwt.version>
+    <gwt.version>2.8.2</gwt.version>
     <!-- TODO test with 2.x version in ip-bom -->
     <hsqldb.version>1.8.0.7</hsqldb.version>
     <javaee.api.version>7.0</javaee.api.version>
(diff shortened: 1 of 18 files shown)


mbarkley:
Going to give this another try.

mbarkley:
Jenkins, please retest this.

psiroky:
The test fails seem related. Unfortunately I don't really know what's going on there. @mbarkley any pointers?

mbarkley:
@psiroky I can take a look tomorrow. But in case you get to it first, the "test already finished" error has to do with some async tests which have to explicitly call `finish` for the test to end. In the past when I have seen this error, it's been from a callback from a previous test that wasn't cleaned up.

An easy way to see if the problem is state contamination between tests is to run one of the failing tests individually and see if the problem persists.

psiroky:
@mbarkley the tests indeed pass when executed individually. I've spent some time trying to debug this, but it's far more complex than I expected. I would leave this to you if you don't mind...

mbarkley:
@psiroky So the most immediate cause of failure is [this](https://github.com/gwtproject/gwt/commit/d9fe18975d57c6cafa56ae2c813f0fee58445c4a#diff-1dcef66e678f6b3e9d5669f53575da23L95).

I've verified with some of the tests that adding this to the relevant gwt.xml file fixes the issue.

After adding a lot of print statements, I haven't the faintest idea how we are calling finish multiple times, but it appears that has always been the case and with this property set to `IGNORE`, it did not stop test execution.

Can you please update your PR by sprinkling in this magic property to the relevant gwt.xml files?

psiroky:
Thanks for the details, I will take a look. Should I add this property to the main `gwt.xml` files (located in the `src/main/java`) or better to add it just for the tests (inside `src/test/java`)?

mbarkley:
Let's start with just the src/test gwt.xml files for now.

psiroky:
@mbarkley this should be ready to go. We could wait for the other repos to get the same upgrade (PRs are created, waiting for successful downstream build), but I think it is not strictly needed (e.g. all the repos can be upgraded individually without any issue).

PR #314, opened 2017-11-21
[ERRAI-1083] allow @Property without a value like disabled, checked a…

treblereel:
…nd selected.
JIRA: https://issues.jboss.org/browse/ERRAI-1083


treblereel:
yeap, it's about convenience. I would like to add that setPropertyBoolean(java.lang.String name, boolean value) and methods like that act the same. So it's not a big deal, but it could save time and, from my point of view, it's looks natural coz value="true" is not necessary.

mbarkley:
Yeah, IMO the convenience of not having to specify the `value = "true"` for some properties is out-weighed by the fact that it opens the door to forgetting to specify a value.

Sorry, but I won't be merging this change.

PR #324, opened 2017-12-16
Ignore bindable interfaces

BenDol:
Sometimes interfaces are picked up in the bindable wildcard check. We can ignore them as we get an exception regarding no default constructor existing.


tiagobento:
ok to test

mbarkley:
Thanks @BenDol!

PR #325, opened 2017-12-16
Gracefully ignore non-public properties in bindable proxies

BenDol:
Also log a warning.

There were cases where I couldn't make complex classes bindable due to them having private properties.

Note that if there are use cases where protected property getters/setters need to be bound, it might be that we need to consider adding a check for protected since the `deepUnwrap` or `unwrap` access the getter/setters methods that throw an error.


tiagobento:
ok to test

mbarkley:
Thanks @BenDol!

PR #331, opened 2018-02-22
ERRAI-1099: Update current pages state without performing a full page navigation

BenDol:
https://issues.jboss.org/browse/ERRAI-1099

This is something that we find very useful when it comes to retaining our pages workflow in the URL. With this we just call `updateState` with a new set of state data (or a modified version) and we retain the pages initial state.

Let me know if there are any unseen limitations here or if there would be a better way to manage something like this, but as far as I have tested it seems to be working very well.

Note that this isn't actually invoking a navigation, it is simply updating the URL for the next navigation attempt. If this isn't done the states will always match the currentPageToken's states when, navigating backwards for example and therefore the navigation process is ended prematurely.

This introduces the `@PageUpdate` lifecycle hook that will be invoked when Navigation.updateState is called.

Tests provided.

Thanks!

### DIFF
--- a/errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageUpdate.java
+++ b/errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageUpdate.java
@@ -48,9 +48,7 @@ import java.lang.annotation.*;
  * @see Page
  * @see PageState
  * @see Navigation
- * @see PageShowing
- * @author Daniel Sachse <mail@w0mb.at>
- * @author Jonathan Fuerth <jfuerth@redhat.com>
+ * @see PageShown
  * @author Ben Dol
  */
 @Target(ElementType.METHOD)
--- a/errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageUpdate.java
+++ b/errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageUpdate.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright (C) 2018 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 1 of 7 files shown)


tiagobento:
ok to test

BenDol:
@mbarkley what would be the easiest way to do this?

mbarkley:
Well, one option would be to call the generated `pageShown` method for the current page. This updates the `@PageState` and calls the `@PageShown` method. This also gives a page a chance to react to the navigation.

The more complex option is to create a new generated method on `PageNode` so that the state can be set without calling the `@PageShown`.

I think calling the `@PageShown` method might be a good idea, but WDYT?

BenDol:
I'll test it out by invoking the `pageShown`, thanks for the feedback.

BenDol:
@mbarkley I have decided the best way to manage this is with a new lifecycle state that we can use for `updateState` calls. This way we are properly differentiating between actual navigation `shown` and this simple state update.

I have provided 2 simple tests that should pass (they did when I tested them).

Thanks!

errai-navigation/src/main/java/org/jboss/errai/ui/nav/client/local/PageUpdate.java
```diff
@@ -0,0 +1,61 @@
...
+ * The target method can have any access type: public, protected, default, or
+ * private.
+ * <p>
+ * If the target method throws an exception when called, behaviour is undefined.
+ *
+ * @see Page
+ * @see PageState
+ * @see Navigation
+ * @see PageShowing
+ * @author Daniel Sachse <mail@w0mb.at>
```
mbarkley: You can remove these other author tags from this annotation.

PR #332, opened 2018-03-02
Implement PayloadPreprocessor for server side payload processing.

BenDol:
This allows us to preprocess the payload before it is encoded.

The most predominant use case we have at the moment is this example:
```java
public class HibernatePayloadPreprocessor implements PayloadPreprocessor {

    @Override
    public void process(Map<String, Object> payload) {
        if(payload.containsKey("MethodReply")) {
            payload.put("MethodReply", HibernateUtil.cleanFromProxies(payload.get("MethodReply")));
        }
    }
}
```
Then we would just need to call:
```java
ErraiProtocolServer.addPreprocessor(new HibernatePayloadPreprocessor());
```

This allows us to preprocess the payload before being sent to the encoder. We are able to avoid: `org.hibernate.LazyInitializationException: could not initialize proxy - no Session` exceptions.

Thanks!


mbarkley:
ok to test

BenDol:
See https://github.com/errai/errai/pull/334

PR #336, opened 2018-03-27
Fixed double slashes in Jaxrs path

jschoedt:
```
RestClient.setApplicationRoot("/root"); 
//and
RestClient.setApplicationRoot("/root/");
```

Both creates a path like "/root//endpoint"

### DIFF
--- a/errai-jaxrs/errai-jaxrs-client/src/main/java/org/jboss/errai/enterprise/client/jaxrs/api/RestClient.java
+++ b/errai-jaxrs/errai-jaxrs-client/src/main/java/org/jboss/errai/enterprise/client/jaxrs/api/RestClient.java
@@ -183,8 +183,6 @@ public class RestClient {
 
     Assert.notNull(remoteService);
     Assert.notNull(callback);
-    if (baseUrl != null && !baseUrl.endsWith("/"))
-      baseUrl += "/";
 
     T proxy = proxyProvider.getRemoteProxy(remoteService);
 
@@ -210,9 +208,6 @@ public class RestClient {
       return ""; 
     } 
     else {
-      if ($wnd.erraiJaxRsApplicationRoot.substr(-1) !== "/") {
-        return $wnd.erraiJaxRsApplicationRoot + "/";
-      }
       return $wnd.erraiJaxRsApplicationRoot;
     }
   }-*/;


mbarkley:
ok to test

tiagobento:
ok to test

jschoedt:
I would like to fix the test but I can not access the error details and 'mvn test' runs fine for me locally...

tiagobento:
Hi @jschoedt, thanks for the PR. Looks like `org.jboss.errai.enterprise.jaxrs.client.test.ConfigurationTest.testApplicationRootWithMissingSlash` is failing due to an assertion error. Is this test passing on your local build?

See the error:
```
Error Message
    expected: </root/>, actual: </root>

Stacktrace
    junit.framework.AssertionFailedError: expected: </root/>, actual: </root>
	at org.jboss.errai.enterprise.jaxrs.client.test.ConfigurationTest.testApplicationRootWithMissingSlash(ConfigurationTest.java:55)
```

mbarkley:
@jschoedt Because our test suite can take a while to run, it is hidden behind a maven profile. We typically run `mvn clean install -Pintegration-test`. You can use `-Dit.test=TestClassName` to run a specific test class.

jschoedt:
I am having issues running `mvn clean install -Pintegration-test`. If I run it on master (locally without my changes) I get:

```
[ERROR] Tests run: 127, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 43.085 s <<< FAILURE! - in org.jboss.errai.enterprise.jaxrs.client.test.AllJaxrsTests
[ERROR] testGetWithDatePathParam(org.jboss.errai.enterprise.jaxrs.client.test.PathParamIntegrationTest)  Time elapsed: 0.136 s  <<< ERROR!
```

 
Also running a single test does not work for me. Would this be the command?
`mvn clean install -Pintegration-test -Dit.test=org.jboss.errai.enterprise.jaxrs.client.test.ConfigurationTest`

To fix the test I would remove it entirely since the suffix of the path (the api path without the root) already has a starting '/' in my case. Adding the '/' to the root would only be necessary if the suffix does not have a starting '/'. I am not aware if that can ever happen? If not then the code should instead remove any tailing '/' from the root path. Correct?

tiagobento:
@jschoedt Hi, sorry for the late response.

When you say "Both creates a path like `"/root//endpoint"`", what is your config? Where does `endpoint` come from on this example?

kiereleaseuser2:

kiereleaseuser2:

jschoedt:
@tiagobento  sorry for the late reply. I have been on vacation :-)
Yes 'endpoint' comes from eg:

    @GET
    @Path("/projects")
    @Produces({"application/json"})
    @ApiOperation(value = "", tags = {"project",})
    @ApiResponses(value = {
            @ApiResponse(code = 200, message = "project response", response = ProjectResult.class),
            @ApiResponse(code = 200, message = "unexpected error", response = ErrorModel.class)})
    ProjectResult getAllProjects();

I tried with `@Path("projects")` as well, but not luck. Both produce a request like: /api/v1//projects

It appears that some web servers will disregard the extra / but others will fail.

tiagobento:
@jschoedt No worries, I haven't been able to give you the fastest replies lately as well :-)

_tl;dr_

I finally was able to test it locally but no luck with reproducing it. I used the errai-jaxrs-demo-crud example of this repo. **It would be very helpful if you could put together an example project or send me some more detailed resources to look at.**

***

Here's a more detailed explanation of how requests are made when using Errai JAX-RS.

For every `@Path`-annotated interface, Errai will generate an `AbstractJaxrsProxy` implementation inside a file called `JaxrsProxyLoaderImpl`. You can search for it on the .errai folder during development. It implements the methods declared on the `@Path`-annotated interface and does the AJAX stuff for you. 

For the `CustomerService.listAllCustomers` method, the generated method is as follows:

```java
public List listAllCustomers() {
        StringBuilder url = new StringBuilder(getBaseUrl());
        url.append("rest/customers");
        RequestBuilder requestBuilder = new RequestBuilder(RequestBuilder.GET, url.toString());
        requestBuilder.setHeader("Accept", "application/json");
        sendRequest(requestBuilder, null, new ResponseDemarshallingCallback() {
          public Object demarshallResponse(Response response) {
            if (response.getStatusCode() == 204) {
              return null;
            } else {
              return MarshallingWrapper.fromJSON(response.getText(), List.class, Customer.class);
            }
          }
        });
        return null;
      }
```

Important here is the second line, where you can see that the `getBaseUrl()` return gets concatenated with the `"rest/customers"` string. That worries me, because if we remove the trailing "/" from the application root, this URL will be invalid, because the `getBaseUrl()` implementation looks like this:
https://github.com/errai/errai/blob/master/errai-jaxrs/errai-jaxrs-client/src/main/java/org/jboss/errai/enterprise/client/jaxrs/AbstractJaxrsProxy.java#L75

Also, that made me wonder how do you end up with a double-slashed URL when there's no slash at the beginning of the path? The only way I could do it was by calling `RestClient.setApplicationRoot("/something//")`.

There's a "problem" with double slashed if you include them manually on your `@Path` annotations, but then I think it _might_ be intentional, so no need in changing that. See how Errai handles `@Path` annotations values: https://github.com/errai/errai/blob/master/errai-jaxrs/errai-jaxrs-client/src/main/java/org/jboss/errai/enterprise/rebind/JaxrsResourceMethod.java#L42

Please tell me if I'm missing something..

jschoedt:
@tiagobento Thank you for the insights. Now I was able to debug it some more.

I think I have identified a problem. With my interface looking like this (minimal example):

```
@Path("")
@Api(value = "", description = "")
public interface ProjectApi {

    @POST
    @Path("projects")
    @Produces({"application/json"})
    @ApiOperation(value = "", tags = {"project",})
    @ApiResponses(value = {
            @ApiResponse(code = 200, message = "project  response", response = Project.class),
            @ApiResponse(code = 200, message = "unexpected error", response = ErrorModel.class)})
    Project addProject(@Valid Project project);
}
```


It generates this on my machine:

```
      public Project addProject(final Project a0) {
        StringBuilder url = new StringBuilder(getBaseUrl());
        url.append("/projects");
        RequestBuilder requestBuilder = new RequestBuilder(RequestBuilder.POST, url.toString());
        requestBuilder.setHeader("Accept", "application/json");
```


Again my base path is: `/api/v1`

So adding the extra "/" to the base path will create the double "//".

Can you check if you get an other result?

An other problem I see is that I think it is totally fine to have an interface like this:

```
@Path("/")
@Api(value = "", description = "")
public interface ProjectApi {

    @POST
    @Path("projects")
    @Produces({"application/json"})
    @ApiOperation(value = "", tags = {"project",})
    @ApiResponses(value = {
            @ApiResponse(code = 200, message = "project  response", response = Project.class),
            @ApiResponse(code = 200, message = "unexpected error", response = ErrorModel.class)})
    Project addProject(@Valid Project project);
}
```
or like this
```
@Path("")
@Api(value = "", description = "")
public interface ProjectApi {

    @POST
    @Path("/projects")
    @Produces({"application/json"})
    @ApiOperation(value = "", tags = {"project",})
    @ApiResponses(value = {
            @ApiResponse(code = 200, message = "project  response", response = Project.class),
            @ApiResponse(code = 200, message = "unexpected error", response = ErrorModel.class)})
    Project addProject(@Valid Project project);
}
```

In both cases that will create the double slashes if the extra "/" is automatically added to the base path right? Maybe there is some reason to add the extra "/" but IMO it would be better to leave it to the user to add the tailing "/" to the base path if he needs it. Otherwise I think the code should check for double slashes and only add it if needed.

kiereleaseuser2:
Build finished. 2762 tests run, 7 skipped, 1 failed.

mareknovotny:
Jenkins retest this please

mareknovotny:
outdated PR, so closing now

PR #337, opened 2018-03-29
ERRAI-1104: Ability to intercept any remote call from the ClientMessageBus

BenDol:
Allows us to do the following as an example:
```java
public class PingMessageInterceptor implements MessageInterceptor {

    private long begin;

    @Override
    public void beforeCall(Message message) {
        begin = new Date().getTime();
    }

    @Override
    public void afterCall(Message message) {
        // check the ping and perform tasks
    }
}
```

```java
ErraiBus.get().addInterceptor(interceptor);
```
Now we can make checks on any remote call rather than just specific calls.

### DIFF
--- a/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
+++ b/errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
@@ -16,46 +16,21 @@
 
 package org.jboss.errai.bus.client.framework;
 
-import static org.jboss.errai.bus.client.protocols.BusCommand.RemoteSubscribe;
-import static org.jboss.errai.bus.client.protocols.BusCommand.RemoteUnsubscribe;
-import static org.jboss.errai.bus.client.util.BusToolsCli.isRemoteCommunicationEnabled;
-import static org.jboss.errai.common.client.protocols.MessageParts.PriorityProcessing;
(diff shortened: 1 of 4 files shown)


tiagobento:
ok to test

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
```diff
@@ -830,6 +814,18 @@ else if (!delivered) {
     }
   }
 
+  public void signalInterceptorsBefore(Message message) {
+    for(MessageInterceptor interceptor : interceptors) {
```
mbarkley: Please fix the formatting on this loop (`for (...)` not `for(...)`).

errai-bus/src/main/java/org/jboss/errai/bus/client/framework/ClientMessageBusImpl.java
```diff
@@ -830,6 +814,18 @@ else if (!delivered) {
...
+      interceptor.beforeCall(message);
+    }
+  }
+
+  public void signalInterceptorsAfter(Message message) {
+    for(MessageInterceptor interceptor : interceptors) {
```
mbarkley: Please fix the formatting on this loop (`for (...)` not `for(...)`).

mbarkley:
Merged. Thanks @BenDol.

PR #338, opened 2018-04-02
Remove unneeded log.warning(...)

hrstoyanov:
This is a very simple patch that removes a single line of logging code.
```Navigation.setErrorHandler(....)``` allows the user to install custom error handler, which is supposed to handle navigation exception. Unfortunately, there is an extra line of logging code that emits a warning *before* the handler get a chance to process the exceptions. Further, ```DefaultNavigationErrorHandler``` already has plenty of logging already during the exception handling, so warning logs seems duplicate and clutters the browser JS console.

Please remove the single logging line with this pull request.


mbarkley:
ok to test

PR #340, opened 2018-05-09
ERRAI-1108: Interrupt navigation without enforcing a redirect

BenDol:
This valid for onHiding, when we want to cancel the hiding of a page.


mbarkley:
ok to test

PR #346, opened 2018-06-04
ERRAI-1110: Implement 'single only' event support.

BenDol:
This will allow us to setSingleOnlyEventType marking an event type with a single handler in some edge cases.

### DIFF
--- a/errai-cdi/errai-cdi-shared/src/main/java/org/jboss/errai/enterprise/client/cdi/api/CDI.java
+++ b/errai-cdi/errai-cdi-shared/src/main/java/org/jboss/errai/enterprise/client/cdi/api/CDI.java
@@ -74,6 +74,7 @@ public class CDI {
 
   private static Map<String, List<AbstractCDIEventCallback<?>>> eventObservers = new HashMap<>();
   private static Set<String> localOnlyObserverTypes = new HashSet<>();
+  private static Set<String> singleOnlyObserverTypes = new HashSet<>();
   private static Map<String, Collection<String>> lookupTable = Collections.emptyMap();
   private static Map<String, List<MessageFireDeferral>> fireOnSubscribe = new LinkedHashMap<>();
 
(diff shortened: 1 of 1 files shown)


BenDol:
Actually some internal discussion is still needed on our side for this implementation, any input to the implementation is welcome.

tiagobento:
ok to test

tiagobento:
Hi @BenDol, sorry for taking so long to look into the PRs. I'm trying to get back on track gradually.

As far as I understood you're rejecting subscriptions to an Event of a given type if there's already someone listening to it, is that correct? 

If so, I think we're creating a break of expectation on the declarative approach of Events, because although you have stated that a method will receive the Event, in reality it will only do so if no one method has subscribed to that Event dynamically..

Can I ask about your use case so I can understand it a little bit better?

BenDol:
Hi @tiagobento thanks for getting back to me on all the PR's.

So the way I decided to approach the issue we had was at the registration level, perhaps there is a better approach doing it at the delivery of the event. This concept isn't being used by us internally yet, but is something that we wanted to start on now and get feedback on. 

Basically we want to ensure certain events only ever have **one** listener in certain cases. This is helpful when the application has a lot of events and we want to isolate events. having had some time to think about it, it probably makes a lot more sense to do it at the delivery of the event since we should be able to toggle this "flag" on and off when required.

tiagobento:
@BenDol Thanks for providing more info about that.

I think I like your idea of checking if the event has more subscribers than it's supposed to at the delivery. Every time we avoid messing with the codegen stuff is a win :)

I'm still not convinced I fully understand the need to have an event that can only be delivered to a single subscriber.. If you could provide an example case it would really help me to have a better picture of what you're proposing.

kiereleaseuser2:

kiereleaseuser2:

kiereleaseuser2:
Build finished. 2762 tests run, 7 skipped, 0 failed.

mareknovotny:
ok to test

PR #348, opened 2018-06-12
Fix an issue with SecurityContextHoldingSingleton loading

BenDol:
There are cases where this dependency isn't loaded as a singleton.


BenDol:
Nevermind I think it was another issue in disguise.

PR #349, opened 2018-06-19
Log throwable if it's not null

yurloc:
Due to the inverted condition, the throwable was actually ignored if it
wasn't null, which is the opposite of what was intended.

### DIFF
--- a/errai-bus/src/main/java/org/jboss/errai/bus/client/util/ErrorHelper.java
+++ b/errai-bus/src/main/java/org/jboss/errai/bus/client/util/ErrorHelper.java
@@ -226,7 +226,7 @@ public class ErrorHelper {
             "\ndisconnect: " + disconnect;
 
     if (!(e instanceof MessageDeliveryFailure && ((MessageDeliveryFailure) e).isRpcEndpointException())) {
-      if (e != null) {
+      if (e == null) {
         logger.error(logMessage);
       }
       else {


tiagobento:
@yurloc Hi Jiri, does that fix a particular issue you found?

BenDol:
I also noticed inside the [ErrorHelper](https://github.com/yurloc/errai/blob/3614a067f8239897361f222c4553d1932845160d/errai-bus/src/main/java/org/jboss/errai/bus/client/util/ErrorHelper.java) there were some strange checks that I wasn't sure about:
* [L162](https://github.com/yurloc/errai/blob/3614a067f8239897361f222c4553d1932845160d/errai-bus/src/main/java/org/jboss/errai/bus/client/util/ErrorHelper.java#L162)
* [L181](https://github.com/yurloc/errai/blob/3614a067f8239897361f222c4553d1932845160d/errai-bus/src/main/java/org/jboss/errai/bus/client/util/ErrorHelper.java#L181)

At face value they don't appear to be doing anything.

tiagobento:
@yurloc Right, I think it's dead code an can be safely deleted. If you want you can amend that to this PR as well.

I just wanted to know if you found those issues during a dev session or are you running some sort of static analysis?

tiagobento:
@BenDol Good catch. Did you find that during a dev session or did you run some sort of static analysis? Feel free to open a PR deleting this code, I think it's dead code and can be safely removed.

BenDol:
Yea it was just while looking through the class, it didn't seem to affect anything for me. The inverted statement that @yurloc is pointing out does have an impact when trying to log the throw-able as you can probably tell though.

I'll just leave the dead code for now, at least until I submit another PR.

yurloc:
@tiagobento Yes it solves the issue that when a MessageDeliveryFailure happened (during dev), I couldn't see the exception stack trace only because the condition was inverted and did something else than what was intended.

PR #354, opened 2018-08-01
ERRAI-1115: Fix content type issues with the StandardAsyncServlet

BenDol:
Since the validateContentType method in HttpPollingHandler expects `application/json` we have to ensure our polling is sent as such.

I have a feeling this is actually an issue with the way we are checking inside the HttpPollingHandler. Since I'm not 100% sure how it is meant to work, I will just fix the servlet.

### DIFF
--- a/errai-bus/src/main/java/org/jboss/errai/bus/server/servlet/StandardAsyncServlet.java
+++ b/errai-bus/src/main/java/org/jboss/errai/bus/server/servlet/StandardAsyncServlet.java
@@ -72,6 +72,7 @@ public class StandardAsyncServlet extends AbstractErraiServlet {
 
     
     final AsyncContext asyncContext = request.startAsync();
+    asyncContext.getResponse().setContentType("application/json");
     asyncContext.setTimeout(60000);
     queue.setTimeout(65000);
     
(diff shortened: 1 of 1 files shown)


tiagobento:
ok to test

bendol-i3:
This might be a work in progress, I need to look into the issues this weekend.

BenDol:
This might be a work in progress, I need to look into the issues this weekend.

tiagobento:
@BenDol Thanks for the PR. Let me know when you think it's ready.

kiereleaseuser2:

kiereleaseuser2:

BenDol:
@tiagobento I wonder if you could help me some issues migrating over to the async servlet, I'm trying to figure out why the `BusLifeCycleListener#busOffline` is no longer called when I switch over to use the `StandardAsyncServlet` from the `DefaultBlockingServlet`. It seems we're no longer maintaining the state of the connection when using this servlet, is that correct? If so, what are all the caveats to using the `StandardAsyncServlet`?

tiagobento:
@BenDol Do you have a reproducer? I'm not that familiar with the internals of Errai Bus TBH.

kiereleaseuser2:
Build finished. 2762 tests run, 7 skipped, 0 failed.

PR #359, opened 2018-09-06
upgraded jboss-ip-bom to 8.1.0.Final

mbiarnes:
jboss-ip-bom 8.1.0.Final is identical to 8.1.0.CR2 - we had to do this as prod will work only with Final versions. In future there won't be anymore CRs for jboss-ip-bom - only Final versions.


kiereleaseuser2:
Build finished. 2761 tests run, 5 skipped, 0 failed.

PR #360, opened 2018-09-12
Fix for ECJ - GWT conflict

wmedvede:
- This commit keeps errai-codgen module aligned with the ecj version used by gwt-dev by doing a shade of the ecj classes contained in the gwt-dev.jar.
      I this way the rest of the WB can use a different ecj version, etc, tipically configured in the droolsjbpm-build-boostrap file."

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ gwt-unitCache/
 class_list.txt
 *.jpage
 .factorypath
+dependency-reduced-pom.xml
--- a/errai-codegen/pom.xml
+++ b/errai-codegen/pom.xml
@@ -42,14 +42,124 @@
     <dependency>
       <groupId>com.google.gwt</groupId>
       <artifactId>gwt-dev</artifactId>
-      <!-- This should not be packaged with the Errai Distro. There are no
-runtime dependencies on it and it breaks deployment on JBoss AS and Tomcat -->
-      <scope>provided</scope>
+      <!-- We need a compile scope for making the shade work (see below), but we don't want to inherit the
+      transitive dependencies of the gwt-dev.jar, so for safety we exclude them manually -->
+      <exclusions>
+        <exclusion>
+          <groupId>com.google.code.findbugs</groupId>
+          <artifactId>jsr305</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.google.code.gson</groupId>
+          <artifactId>gson</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.ow2.asm</groupId>
+          <artifactId>asm</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.ow2.asm</groupId>
+          <artifactId>asm-util</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>org.ow2.asm</groupId>
+          <artifactId>asm-commons</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>colt</groupId>
+          <artifactId>colt</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>ant</groupId>
+          <artifactId>ant</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>commons-collections</groupId>
+          <artifactId>commons-collections</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>com.ibm.icu</groupId>
+          <artifactId>icu4j</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>tapestry</groupId>
+          <artifactId>tapestry</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>net.sourceforge.htmlunit</groupId>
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -35,3 +35,4 @@ gwt-unitCache/
 class_list.txt
 *.jpage
 .factorypath
+dependency-reduced-pom.xml
```
tiagobento: Hi @wmedvede, is this line necessary or did it slip in? :)

wmedvede:
Hello @manstis @porcelli  @ederign @tiagobento 
I believe this is the correct solution for the issue we are having this days after the ECJ version change in droolsjbpm-build-boostrap.

The problem we have is basically that errai is "linked" to a different ECJ compiler version.
Where to errai take the ECJ from? well, basically errai is using ECJ classes coming from gwt-dev.jar file. Surprisingly the ECJ classes are packaged there.

So the fix is more or less to warranty that errai continue using this classes and let the rest of the WB apps use the new ECJ version.
This is done basically with the shade of this commit.

Would you mind take a look and give your thoughts?

Thanks.

Hi @tiagobento consider this a not to merge PR until more people can test it and check, etc.

.gitignore
wmedvede: @tiagobento I think it's necessary since this file is generated by the shade plugin

tiagodolphine:
@wmedvede hi, I'm taking a look on the solution, I was investigating the issue as well, I got one other idea, that basically is to make the `ClassChangeUtil.JDTCompiler.compile(...)` method to use the implementation of `BatchCompiler`  from the **ECJ** instead of using it from the **GWT** implementation. In this way the conflict should be solved.  WDYT?
Just brainstorming about the possible solutions.

Thanks.

wmedvede:
@tiagodolphine it could be a good solution. yes, if we can break this "dependency" to the ECJ classes in the gwt-dev.jar it's probably better. But, my initial thought was that likely it was done this way to be sure the client classes generated with errai uses exactly the same compiler used by GWT. I don't know honestly this internals, but it's the only reason that comes to my mind. I think this way they had some level of confidence that classes produced by gwt are compiled with the same compiler as GWT. And also with some hope, errai automatically bumps up the compiler compiler version when they change the gwt-dev.jar version.
But, yes, I'm happy with whatever solution that can get us rid form this.
I think I need a pizza ;), tomorrow more.

kiereleaseuser2:
Build finished. 2761 tests run, 5 skipped, 0 failed.

manstis:
@wmedvede @tiagobento OK, so let me try to understand the status. At the moment @wmedvede PR includes (shades) GWT's ECJ version in Errai's ```code-gen``` module? What work is @tiagobento actively researching? Is this PR considered short-term (until @tiagobento finishes his investigations)? If we verify the webapps work with this fix are we merging or waiting?

wmedvede:
Hello @manstis, I believe this is more or less the status:

1) ECJ compiler version was modified and merged both in master and 7.11.x branches (probably passing zero tests) causing next all showcases/webapps to fail.

2) both you and me have wasted countless time doing several try&errors by modifying the dependency configurations, etc, along the different showcases, projects

3) more people has also wasted time in the middle, and suggested tricks, etc, that we also tested

4) from 1,2 and 3, we could get a sort of "partial/dirty solution" that leaves some use cases still not working 100% well. But at least seems like we could get rid from the build issues.

5) yesterday night I found this alternative that is under testing/discussion, and that seems to solve the issue.

6) @tiagobento has a different possible solution not yet PR-red, so it can't be tested.

7) I see you are in process of testing 5) and seems to be working well. Additionally to the tests i did Yesterday I'll now check kie-wb-distributions now.

8) the cut date for the 7.11.x branch is today

My opinion is that when we finish local tests, we should go ahead with the fix proposed in this PR

wmedvede:
@manstis forgot to mention, the build for the PR with the partial fix for kie-wb-distributions/7.11.x generated all the artifacts well, I've checked the generated .wars in WF11 locally and worked, BUT there was a selenium test that failed doing the verification lf the ProcessInstancesPerspective load. It's really weird since the change is not related with this perspective at all, and the generated war works well in WF11. Not sure if we have also randomly failing tests at selenium side, but this is the status right now regarding the "partial fix" applied to kie-wb-distributions/7.11.x branch. Will try to run it again...

tiagodolphine:
Hi @wmedvede I tested the solution I proposed but I got some other conflicts after inserting the ECJ on the  errai-codegen so, for now, I think this solution with shades you made is the best and quicker one. If the showcases are working well and there are no side effects my vote is to go on.
After this get merged I think the https://github.com/kiegroup/kie-wb-common/pull/2112 can be closed.
Thanks !

manstis:
@tiagobento All my testing has been successful. Can you approve so we can merge and tidy up?

wmedvede:
@manstis OK, I've reproduced the fix with errai 4.3.x branch and built all WB related 7.11.x repositories, etc., and all builds were ok. Additionally I've verified that kie-wb-distributions for 7.11.x branch generated artifacts are working well.
So from your local tests + my Yesterday's master tests + my today's tests in 7.11.x I believe we can go ahead with this fix both for master, 7.11.x/4.3.x branches.

I'll also create the peer PR for the 4.3.x branch.

I'm trying to add @jomarko to the thread since he's aware of the different threads with the issue, but seems like he's not registered on this repository. Anyway, I'll ping him.

PR #361, opened 2018-09-13
[4.3.x] Fix for ECJ - GWT conflict

wmedvede:
- This commit keeps errai-codgen module aligned with the ecj version used by gwt-dev by doing a shade of the ecj classes contained in the gwt-dev.jar.
      I this way the rest of the WB can use a different ecj version, etc, tipically configured in the droolsjbpm-build-boostrap file."

(cherry picked from commit 207bb3059160314b7b63204da67516405ecba1ea)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,4 @@ gwt-unitCache/
 class_list.txt
 *.jpage
 .factorypath
+dependency-reduced-pom.xml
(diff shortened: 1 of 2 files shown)


wmedvede:
Hi  @manstis @jomarko this is the peer PR for #360 with the ECJ conflict fix.
Would you mind take a look?

Thanks!

PR #362, opened 2018-09-23
Use extends for roles and groups allow generic overrides

BenDol:
This allows us to use our own groups/role domain entities without changing the name of the properties and hacking around it.


kiereleaseuser2:

PR #401, opened 2020-08-18
restored version of maven-deploy-plugin

mbiarnes:


treblereel:
@mbiarnes looks good for me

PR #435, opened 2021-08-11
Reorganize the `main` branch

tiagobento:
This PR reverts untested changes that were made to `main` along with adding the contents of `4.11.x` and `4.14.x` branches, as we need them to be on a new release.

This PR will leave `main` at the `4.15.0-SNAPSHOT` version, meaning that the release process can be followed normally after it's merged.

This PR is the combination of the changes necessary for the Conscious Language Initiative and the WildFly 23 upgrade.

cc @jesuino


errai-bom/pom.xml
```diff
@@ -76,22 +76,17 @@
     <jboss.web.version>7.0.16.Final</jboss.web.version>
     <jsr305.version>1.3.9</jsr305.version>
     <mojo.executor.version>2.2.0</mojo.executor.version>
-    <netty.codec.http.version>4.1.59.Final</netty.codec.http.version>
+    <netty.codec.http.version>4.1.48.Final</netty.codec.http.version>
```
mareknovotny, line 79: I think we should keep this upgrade or do 4.1.63 as it is in Wildfly https://github.com/wildfly/wildfly/blob/23.x/pom.xml#L304
This was originally done due https://github.com/netty/netty/security/advisories/GHSA-5mcr-gq6c-3hq2, so we should not revert this version

errai-bom/pom.xml
```diff
@@ -144,8 +147,8 @@
     <version.org.powermock>2.0.7</version.org.powermock>
     <version.org.quartz-scheduler>2.3.2 </version.org.quartz-scheduler>
     <version.org.slf4j>1.7.26</version.org.slf4j>
-    <version.org.wildfly>23.0.1.Final</version.org.wildfly>
-    <version.org.wildfly.arquillian>3.0.1.Final</version.org.wildfly.arquillian>
+    <version.org.wildfly>23.0.2.Final</version.org.wildfly>
+    <version.org.wildfly.arquillian>2.2.0.Final</version.org.wildfly.arquillian>
```
mareknovotny, line 151: this also should not be reverted from 3.0.1.Final as Wildfly may not work properly with older version. WF uses https://github.com/wildfly/wildfly/blob/23.x/pom.xml#L470 that version.

mbiarnes (approved):
LGTM - as I am not a specialist of what works with what please follow Marek N. comments

tiagobento:
@mareknovotny Done.

mareknovotny (approved):

tiagobento:
Jenkins retest this

PR #485, opened 2026-08-05
Updated to new version 5.0.0-SNAPSHOT

akumar074:


errai-docs/src/main/docbook/en/reference.xml
```diff
@@ -1,24 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE book PUBLIC "-//OASIS//DTD DocBook XML V4.5//EN" "http://www.oasis-open.org/docbook/xml/4.5/docbookx.dtd">
-
+<?asciidoc-toc?>
```
yesamer: @akumar074 Is this file changed intentionally?

akumar074: No, It wasn't. It is generated part of code. I just reverted it. Thanks for notifying this.

yesamer (approved):

mareknovotny:
just curious why new major version? there is no other change in the source code ?

akumar074:
Hi @mareknovotny we have major upgrades with Java 17, Jakarta 10, and wildfly 35 too since 4.18.0.Final. and we are expecting more changes to make errai compatible with modern stack.

